PR #31, opened 2015-01-08
move to a common provider for the server; add a test for the code comple...

devoncarew:
Move to share code between the non-ae server and the ae server. Abstract out the various dependencies. This has the nice side benefit of making the server implementations easily testable.

@lukechurch 

Only the code complete service is moved over. The other 3 will need to be as well. This should result in a lot less (duplicated) code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ To run the server, run:
 
     dart bin/dartpad_server.dart --port 8082
 
-The server will run from port 8082 and export several JSON APIs, like `/api/compile`
-and `/api/analyze`.
+The server will run from port 8082 and export several JSON APIs, like
+`/api/compile` and `/api/analyze`.
 
 ## See also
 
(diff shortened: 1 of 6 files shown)


lukechurch:
I'm going backwards and forwards on this.

I like the testability. That's awesome.

I'm not sure how much we're going to get out of maintaining the non-ae server, and whether it's going to become something that means that we have a bunch of extra complexity without much gain.

E.g. I think we may want to read from the datastore to support ordering of the results for completion. If we do that we're going to also need to mock up the data store, and it's not so obvious how meaningful that would be locally.
On the other hand, we're probably going to need to do most of the work for testability anyway.

So for now, LGTM, but if we find ourselves not using the non-ae server, and it's just adding an abstraction layer, shall we consider removing?

devoncarew:
> So for now, LGTM, but if we find ourselves not using the non-ae server, and it's just adding an abstraction layer, shall we consider removing?

SGTM. I'm not that committed to the non-ae server. I think this change pays for itself in the testability angle. But if we end up using a lot more AE apis, and mocking everything out gets too expensive, we can re-visit.

PR #37, opened 2015-01-09
always send responses back as utf8

devoncarew:
@lukechurch, make sure we always send responses back as utf8. This makes non-ascii responses happy.


devoncarew:
Luke, when you get a chance, can you patch this in and re-push to appengine? Thanks!

lukechurch:
Pushing, it just takes a few minutes

PR #38, opened 2015-01-10
fix cors with safari

devoncarew:
@lukechurch, I think this will fix a cors issue with safari. The site can't analyze or compile - some googling led me to believe that we need to add the allow-control-headers for `Content-Type`.

### DIFF
--- a/lib/dartpad_gae.dart
+++ b/lib/dartpad_gae.dart
@@ -33,10 +33,10 @@ class GaeServer {
   Future start() => runAppEngine(requestHandler);
 
   void requestHandler(io.HttpRequest request) {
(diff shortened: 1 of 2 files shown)


lukechurch:
@devoncarew good catch, thanks. 

ok - I think in this case I'll strip GET and OPTIONs as verbs, since we don't support them on the server anyway, or did your googling suggest they were required?

devoncarew:
No, sounds good - they were copied from a local server example.

devoncarew:
(updated w/ the GET and OPTIONS permissions removed)

lukechurch:
lgtm

PR #40, opened 2015-01-13
work in progress to add package resolution support

devoncarew:
Work in progress PR to add support for packages. This adds functionality to resolve a given set of packages into the transitive closure of referenced packages and specific package versions. It does not yet have support for getting (package,version) contents.


lib/src/pub.dart
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  String get version {
+    ProcessResult result = Process.runSync('pub', ['--version']);
+    return result.stdout.trim();
+  }
+
+  Future<PackagesInfo> resolvePackages(List<String> packages) {
+    Directory tempDir = Directory.systemTemp.createTempSync('temp_package');
+
```
lukechurch: Could we use the named param here i.e. createTempSync(prefix: 'temp_package') -- I had to look up in the docs to see whether this was the name, and therefore was creating a race condition

devoncarew: Sounds good. Will do once I circle back to this.

devoncarew:
@seaneagan, I'm adding a bit of pub functionality to the dartpad backend. It's two bits:

- Given a list of packages, resolve the list to the full transitive set of referenced packages and their versions.
- And, given a package and a version, it can return a local directory with the contents of the package's `lib/` folder (downloading it if necessary in the process).

I looked at `den`, and it wasn't immediately obvious if this functionality would be useful there. In any case, if you can see a match, I'm happy to move some of this functionality over to there and re-consume it from `den`.

devoncarew:
And @lukechurch, ready for review -

lukechurch:
@devoncarew As discussed - lets keep this as a separate branch until we decide to harden and focus on support for packages, this reduces the risk of accidentally exposing something.

lukechurch:
PS - Love the functionality that this should bring in when we do land it.

seaneagan:
@devoncarew thanks for checking in on that!  I think the pubspec.lock
parsing / listing of resolved versions might fight in as a `den list`
command eventually.  I think den is focused on resolving against a complete
pubspec with version constraints instead of a list of package names with
implied `any` constraints though, but maybe that is something you are
adding in the future.  I think I'll sit on it for now, and if I add
something like this to den in the future, I'll sent a PR back here to
consume it from there.

devoncarew:
@seaneagan, sounds good. From talking to Luke we're going to hold off on working on package support in the short term. But `den`-wise, I was wondering if any of the functionality makes sense in den-as-a-library - if there's overlap if what you're trying to do or if they're disjoint. I'd definitely be interested in pushing functionality into it that makes sense; I'd be less interested in consuming den as a cli app (from this project). Hope that makes sense -

devoncarew:
@lukechurch, this has been updated to the latest codebase. I'd like to land this now in order to experiment w/ package: support -

lukechurch:
@devoncarew lets talk offline.

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+ * package+version from pub.dartlang.org and store it in a disk cache. As the
+ * cache is populated the cost of asking for a package's `lib/` folder contents
+ * will go to zero.
+ *
+ */
+class Pub {
+  Directory _cacheDir;
+
+  Pub() {
+    _cacheDir = new Directory(path.join(_userHomeDir(), '.dartpadcache'));
```
lukechurch: Could we put it in a temp directory so that each time the process gets reset it gets rebuilt again. This will help bound the maximum length of the potential for any problems.

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+}
+
+/**
+ * A packake name and version tuple.
+ */
+class PackageInfo {
+  final String name;
+  final String version;
+
+  PackageInfo(this.name, this.version);
```
lukechurch: Can we put some form of safe-string sanitisation on these? e.g. alpha numerics + _ only?

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+
+      // Run pub.
+      ProcessResult result = Process.runSync(
+          'pub', ['get'], workingDirectory: tempDir.path);
+      if (result.exitCode != 0) {
+        if (result.stderr.isNotEmpty) {
+          return new Future.value(result.stderr);
+        } else {
+          return new Future.value(
+              'failed to get pub packages: ${result.exitCode}');
```
lukechurch: Lets put in a log call for these

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+    return new Future.value();
+  }
+
+  /**
+   * Return the directory on disk that points to the `lib/` directory of the
+   * version of the specific package requested.
+   */
+  Future<Directory> getPackageLibDir(PackageInfo packageInfo) {
+    try {
+      String dirName = '${packageInfo.name}-${packageInfo.version}';
```
lukechurch: Needs sanitisation, this could be put in the package info data objects though.

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+      Directory target) {
+    final String base =
+        'https://storage.googleapis.com/pub.dartlang.org/packages';
+
+    // tuneup-0.0.1.tar.gz
+    String tgzName = '${package.name}-${package.version}.tar.gz';
+    return http.get('${base}/${tgzName}').then((http.Response response) {
+      // Save to disk (for posterity?).
+      File tzgFile = new File(path.join(cacheDir.path, tgzName));
+      tzgFile.writeAsBytesSync(response.bodyBytes);
```
lukechurch: flush: true?

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+    if (!_cacheDir.existsSync()) _cacheDir.createSync();
+  }
+
+  Directory get cacheDir => _cacheDir;
+
+  /**
+   * Return the current version of the `pub` executable.
+   */
+  String get version {
+    ProcessResult result = Process.runSync('pub', ['--version']);
```
lukechurch: Can we turn this into a method? sync calling process start seems expensive for a property. It's going to need caching otherwise.

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+  Future<PackagesInfo> resolvePackages(List<String> packages) {
+    Directory tempDir = Directory.systemTemp.createTempSync(
+        /* prefix: */ 'temp_package');
+
+    try {
+      // Create pubspec file.
+      File pubspecFile = new File(path.join(tempDir.path, 'pubspec.yaml'));
+      String specContents = 'name: temp\ndependencies:\n' +
+          packages.map((p) => '  ${p}: any').join('\n');
+      pubspecFile.writeAsStringSync(specContents);
```
lukechurch: Do you need to flush with the writeAs...Sync calls, I don't remember.

lib/src/pub.dart
```diff
@@ -0,0 +1,211 @@
...
+
+      // Run pub.
+      ProcessResult result = Process.runSync(
```
lukechurch: Can we put a timeout guard on this?

lukechurch:
@devoncarew PTAL at these comments, I'm ok with some of them being turned into issues to be dealt with at a later date.

I'd really like to put 'safe-string' guards on the package names before we merge this.

Let me know what you think about these. Thanks,

lib/src/pub.dart
devoncarew: done

lib/src/pub.dart
devoncarew: Sounds good. I'll also add some docs here; I think that the caller should really be in charge of sanitizing strings. This site may just to a hard fail -

lib/src/pub.dart
devoncarew: sgtm

lib/src/pub.dart
devoncarew: Let's assume it's already sanitized by the time it gets here.

lib/src/pub.dart
devoncarew: added

lib/src/pub.dart
devoncarew: sounds good

lib/src/pub.dart
devoncarew: added

lib/src/pub.dart
devoncarew: converted to async, and added a 20 seconds timeout

MikeMitterer:
Hi guys - how is the status on this? Dart-Pad is nice but it really misses the possibility of importing packages.

I want to add some live samples for my MDL/Dart package http://styleguide.material-design-lite.pub/

PR #46, opened 2015-02-01
whitelist OPTIONS requests

devoncarew:
Fix https://github.com/dart-lang/dartpad_ui/issues/76.

@lukechurch, can you push a new version of the server? I think this PR will let Safari talk to the backend again.


lukechurch:
Pushed, please take another look

devoncarew:
Yup, fixed :)

PR #57, opened 2015-02-20
add a build task to generate the discovery document from the API

### LINKED ISSUE #54: generate the discovery document as part of the build process

devoncarew:
Fix #54 - generate the discovery document as part of our build.

@lukechurch, @wibling 

@wibling, what I'm doing now is starting a copy of the server, downloading the discovery document from the localhost URL, writing the file out, and killing the server. Would there be a more straightforward way of doing this? I imagine you need the annotations available at runtime to generate the doc. If the server could respond to a `--gen_doc` command-line flag, are there APIs in the `rpc` package the server could call to generate the doc w/o having to have a server running? As in, the server could either be invoked in a way to start up a server (the 99% use case), or invoked such that it just wrote out its discovery doc and exited?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,11 @@ To run the server, run:
 The server will run from port 8082 and export several JSON APIs, like
 `/api/compile` and `/api/analyze`.
 
+## The API
+
+The discovery doc for the server's REST API is available here:
+http://dart-services.appspot.com/api/discovery/v1/apis/dartservices/v1/rest.
+
 ## See also
 
 The [Dart Pad](https://github.com/dart-lang/dart-pad) repo.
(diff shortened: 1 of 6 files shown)


lukechurch:
@devoncarew lgtm

In general I like to keep a very explicit separation between things that are generated and things that are user created. Otherwise we can get oursevles into a mess losing edits by accident.

So could we put it in: doc/generated/dartservices.json?

devoncarew:
> I like to keep a very explicit separation between things that are generated

Makes sense. I'm hoping that we'll soon have a generated dart library co-located with this discovery document. So, `dartservices.json` and `dartservices.dart`, or some similar name. In `doc/generated`, `doc/`, `generated/`?

wibling:
Hi Devon,

If you want to do it without spinning up the server in a separate process you could do something like this change: https://codereview.chromium.org/948873002/

devoncarew:
Thanks! I'll update this PR once https://codereview.chromium.org/948873002/ lands.

devoncarew:
Merging - I think the build failure (VM segfault) is intermittent.

PR #67, opened 2015-02-23
use tuneup to check all source for analysis issues; fix analysis issues

devoncarew:
- check all source for analysis issues, not just a few specific libraries (use tuneup for this)
- fix 5 outstanding analysis issues
- fix an issue with checking for the `LIBRATO_USER` env token

Issues were (along w/ some fixed in a previous PR):
```
[info] The value of the local variable 'pos' is not used at example/apitest.dart, line 54.
[info] The value of the local variable 'pos' is not used at example/apitest.dart, line 72.
[info] The value of the local variable 'foo' is not used at example/index.dart, line 12.
[info] The value of the local variable 'result' is not used at lib/src/bench.dart, line 125.
[info] The value of the local variable 't' is not used at tool/load_driver.dart, line 29.
```

### DIFF
--- a/example/apitest.dart
+++ b/example/apitest.dart
@@ -51,7 +51,6 @@ void setupComplete() {
   offsetElement.text = 'offset ${_getOffset(editor)}';
 
   editor.onCursorActivity.listen((_) {
-    Position pos = editor.getCursor();
     offsetElement.text = 'offset ${_getOffset(editor)}';
   });
 }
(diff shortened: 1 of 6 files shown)


devoncarew:
@lukechurch, TBR; this should protect us from analysis regressions in the source base. The previous technique only checked a few libraries.

lukechurch:
lgtm, thanks

Issue #51, opened 2015-02-14
Support for async/await?


lukechurch:
Verify support level for async/await
If needs be, rev the compiler and the analyzer to versions that include support

PR #74, opened 2015-03-09
support async await

devoncarew:
- Upgrade the analyzer and dart2js versions to capture support for async await (fix https://github.com/dart-lang/dart-services/issues/51).
- add async tests for both the analyzer and dart2js
- fix some API deltas

@lukechurch


lukechurch:
lgtm, sadly this is going to cause a minor upmerge problem with code completion version, but it shouldn't be too bad.

I'll push this when I do a server upgrade later today.

lukechurch:
PS. Thanks for keeping these two packages moving forwards Devon.

PR #75, opened 2015-03-10
fix an exception with logging

devoncarew:
- use the `cli_util` package to locate the SDK instead of `grinder`
- fix an exception with setting the logging level in the non-AE version of the server

@lukechurch


lukechurch:
I'll look into using the cli scheme. GAE is a bit sensitive to the use of the command lines.

but lgtm

PR #79, opened 2015-03-18
Code completion

lukechurch:
Bring code completion service prototype and the analysis server into master. The details of the API format will probably still change over time.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,4 +1,4 @@
-version: dbtext
+version: cc
 runtime: custom
 vm: true
 api_version: 1
--- a/lib/services_gae.dart
+++ b/lib/services_gae.dart
@@ -16,12 +16,16 @@ import 'package:memcache/memcache.dart';
 import 'package:rpc/rpc.dart' as rpc;
 import 'src/common_server.dart';
 
+import 'src/completer_driver.dart' as completer;
+
+
 const String _API = '/api';
 
 final Logger _logger = new Logger('gae_server');
 
 void main() {
   GaeServer server = new GaeServer('/usr/lib/dart');
+  completer.setup();
 
   // Change the log level to get more or less detailed logging.
   ae.useLoggingPackageAdaptor();
@@ -56,6 +60,7 @@ class GaeServer {
         'POST, OPTIONS');
     request.response.headers.add('Access-Control-Allow-Headers',
         'Origin, X-Requested-With, Content-Type, Accept');
+
     // Explicitly handle an OPTIONS requests.
     if (request.method == 'OPTIONS') {
       var requestedMethod =
@@ -80,8 +85,13 @@ class GaeServer {
       // a plain text handler if we want to support that.
       var apiRequest = new rpc.HttpApiRequest.fromHttpRequest(request, _API);
       apiServer.handleHttpApiRequest(apiRequest)
-          .then((rpc.HttpApiResponse apiResponse) =>
-            rpc.sendApiResponse(apiResponse, request.response))
+          .then((rpc.HttpApiResponse apiResponse) {
+
+            print ("rpc.HttpApiResponse");
+            print (apiResponse.toString());
+
+            return rpc.sendApiResponse(apiResponse, request.response);
+      })
           .catchError((e) {
             // This should only happen in the case where there is a bug in the
             // rpc package. Otherwise it always returns an HttpApiResponse.
--- a/lib/src/analysis_server_server.dart
+++ b/lib/src/analysis_server_server.dart
@@ -1,0 +1,13 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analysis_server/starter.dart';
+
+/**
+ * Create and run an analysis server.
+ */
+void main(List<String> args) {
+  ServerStarter starter = new ServerStarter();
+  starter.start(args);
+}
--- a/lib/src/completer_driver.dart
+++ b/lib/src/completer_driver.dart
@@ -1,0 +1,481 @@
+library services.completer_driver;
+
+import 'dart:io' as io;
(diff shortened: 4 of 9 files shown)


app.yaml
```diff
@@ -1,4 +1,4 @@
-version: dbtext
+version: cc
```
devoncarew: If we're merging this into master, we should update this to the stable (or dev?) appengine version.

lib/services_gae.dart
```diff
@@ -81,8 +86,13 @@ class GaeServer {
...
       apiServer.handleHttpApiRequest(apiRequest)
-          .then((rpc.HttpApiResponse apiResponse) =>
-            rpc.sendApiResponse(apiResponse, request.response))
+          .then((rpc.HttpApiResponse apiResponse) {
+
+            print ("rpc.HttpApiResponse");
+            print (apiResponse.toString());
+
+            return rpc.sendApiResponse(apiResponse, request.response);
+      })
```
devoncarew: indent?

lib/services_gae.dart
```diff
@@ -81,8 +86,13 @@ class GaeServer {
       // a plain text handler if we want to support that.
       var apiRequest = new rpc.HttpApiRequest.fromHttpRequest(request, _API);
       apiServer.handleHttpApiRequest(apiRequest)
-          .then((rpc.HttpApiResponse apiResponse) =>
-            rpc.sendApiResponse(apiResponse, request.response))
+          .then((rpc.HttpApiResponse apiResponse) {
+
+            print ("rpc.HttpApiResponse");
```
devoncarew: Are the prints intended here? We should probably be logging, or remove the prints.

lib/src/analysis_server_server.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
devoncarew: 2015

lib/src/analysis_server_server.dart
```diff
@@ -0,0 +1,13 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analysis_server/starter.dart';
+
+/**
+ * Create and run an analysis server.
+ */
+void main(List<String> args) {
```
devoncarew: Did you mean to have a `main()` method in the `lib/` dir? Generally I believe entrypoints in lib/ should be avoided.

lib/src/completer_driver.dart
```diff
@@ -0,0 +1,479 @@
...
+
+Future<ServerGetVersionResult> sendServerGetVersion() {
+  return server.send("server.getVersion", null)
+      .then((result) {
+    ResponseDecoder decoder = new ResponseDecoder(null);
+    return new ServerGetVersionResult.fromJson(decoder, 'result', result);
+  });
+}
+
+Future sendCompletionGetSuggestions(String file, int offset) {
```
devoncarew: If this is returning a result, we should add that as a type parameter.

lib/src/completer_driver.dart
```diff
@@ -0,0 +1,479 @@
...
+  var params = new CompletionGetSuggestionsParams(file, offset).toJson();
+  return server.send("completion.getSuggestions", params)
+      .then((result) {
+    ResponseDecoder decoder = new ResponseDecoder(null);
+    return new CompletionGetSuggestionsResult.fromJson(decoder, 'result', result);
+  });
+}
+
+Future sendAddOverlay(String file, String contents) {
```
devoncarew: ditto re: returned future result

lib/services_gae.dart
```diff
@@ -23,6 +26,7 @@ final Logger _logger = new Logger('gae_server');
 void main() {
   // TODO: Use the cli_util package to locate the SDK? Or depend on a CLI arg?
   GaeServer server = new GaeServer('/usr/lib/dart');
+  completer.setup();
```
devoncarew: Instead of using top-level functions in the `completer_driver` library, consider having a `CompleterDriver` class. That can wrap all knowledge of the completion service and the embedded analysis server. `lib/common_srver.dart/CommonServer` could then create a singleton instance of the CompleterDriver class.

pubspec.yaml
```diff
@@ -6,7 +6,13 @@ homepage: https://github.com/dart-lang/dart-services
...
   sdk: '>=1.0.0 <2.0.0'
 dependencies:
-  analyzer: ^0.24.0
+  analysis_server:
+    git: git://github.com/lukechurch/dart-analysis-server-temp-working.git
+  analyzer:
+    git: git://github.com/lukechurch/dart-analyzer-temp-working.git
+  dart_style:
+    git: git://github.com/lukechurch/dart_style.git
+#  analyzer: '>=0.22.4 <0.23.0'
```
devoncarew: @lukechurch, I'm going to create a tracking issue to move back to hosted deps. Looks like you had to go through a lot of pain to get this working :(

app.yaml
lukechurch: For this services I'm currently using a different naming scheme with each new name is based on features. I'll update the wiki about this.

lib/services_gae.dart
lukechurch: Will do. Thanks

lib/services_gae.dart
lukechurch: No, I'll remove them.

lib/src/analysis_server_server.dart
lukechurch: Thanks yeah, in this specific case it's because of the way in which we start the Analysis Server process. I'll add a comment.

pubspec.yaml
lukechurch: Yup. Thanks, I'd like to get these back to hosted deps again. That should happen in the 1.10 time frame.

lib/src/completer_driver.dart
lukechurch: yup, done, thanks

PR #83, opened 2015-03-19
add types to completion code

devoncarew:
- add additional types to the completion code
- run our tests in checked mode

@lukechurch

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -13,7 +13,7 @@ import 'package:rpc/rpc.dart';
 import 'analyzer.dart';
 import 'compiler.dart';
 
-import 'completer_driver.dart' as completer;
+import 'completer_driver.dart' as completer_driver;
 
 final Duration _standardExpiration = new Duration(hours: 1);
 final Logger _logger = new Logger('common_server');
@@ -45,11 +45,10 @@ class CompileResponse {
 }
 
 class CompleteResponse {
-  String result;
+  final String result;
 
-  CompleteResponse(List<String> completions) {
+  CompleteResponse(Iterable<String> completions) :
     result = completions.join(",");
-  }
 }
 
 class DocumentResponse {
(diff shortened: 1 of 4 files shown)


lib/src/common_server.dart
```diff
@@ -13,7 +13,7 @@ import 'package:rpc/rpc.dart';
 import 'analyzer.dart';
 import 'compiler.dart';
 
-import 'completer_driver.dart' as completer;
+import 'completer_driver.dart' as completer_driver;
```
devoncarew: renamed this, as seeing variables named `completer` definitely make me think that the object was an instance of the `Completer` class.

lib/src/completer_driver.dart
```diff
@@ -31,8 +31,8 @@ typedef void NotificationProcessor(String event, params);
 Stream<bool> analysisComplete;
 StreamController<bool> _onServerStatus;
 
-Stream completionResults;
-StreamController _onCompletionResults;
+Stream<Map> completionResults;
```
devoncarew: Added types, to help me trace through the code.

lib/src/common_server.dart
```diff
@@ -45,11 +45,10 @@ class CompileResponse {
 }
 
 class CompleteResponse {
-  String result;
+  final String result;
 
-  CompleteResponse(List<String> completions) {
+  CompleteResponse(Iterable<String> completions) :
```
devoncarew: The call to this ctor would have failed if the server was running in check mode (List was expected, but an iterable was passed in). I changed our tests to run in check mode to help catch this. We should also have an integration test for the completion service.

lukechurch: We do have, just as discussed they aren't integrated yet. I suspect they are however in unchecked mode, I'll see and fix that

lib/src/common_server.dart
lukechurch: sg.

lukechurch:
lgtm, thanks!

PR #85, opened 2015-03-19
return all the completion data available as a map

devoncarew:
@lukechurch 

Change the /completions endpoint to return a list of maps rather than a list of strings. This includes all the completion data available from the analysis server. We'll need to update the UI fairly soon after this lands :)

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -45,10 +45,23 @@ class CompileResponse {
 }
 
 class CompleteResponse {
-  final String result;
+  final List<Map<String, String>> completions;
 
-  CompleteResponse(Iterable<String> completions) :
-    result = completions.join(",");
+  CompleteResponse(List<Map> completions) :
+    this.completions = _convert(completions);
+
+  /**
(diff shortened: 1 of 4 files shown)


lib/src/common_server.dart
```diff
@@ -45,10 +45,23 @@ class CompileResponse {
...
+    this.completions = _convert(completions);
+
+  /**
+   * Convert any non-string values from the contained maps.
+   */
+  static List<Map<String, String>> _convert(List<Map> list) {
+    return list.map((m) {
+      Map newMap = {};
+      for (String key in m.keys) {
+        newMap[key] = '${m[key]}';
```
lukechurch: Is this efficient as a way of stringifying the values?

lukechurch:
Thanks @devoncarew, I'll merge this in, it's going to cause the tests to regress, which is fine.

lib/src/common_server.dart
devoncarew: It does more work than it needs to for string values; don't know that it's really a performance concern however -

PR #86, opened 2015-03-21
return the replacement offset and length

devoncarew:
@lukechurch 

- return the replacementOffset and replacementLength for code completion results
- escape Maps and Lists in the cc results, instead of converting to strings. In the future we could move to a more structured type for individual cc results.
- ws changes and re-gen the discovery library

### DIFF
--- a/lib/src/completer_driver.dart
+++ b/lib/src/completer_driver.dart
@@ -126,6 +126,7 @@ void dispatchNotification(String event, params) {
     _onServerStatus.add(true);
   }
 
+  // TODO: Should we be ignoring the ones that aren't marked 'isLast'?
   if (event == "completion.results" && params["isLast"]) {
     _onCompletionResults.add(params);
   }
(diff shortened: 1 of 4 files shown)


lib/src/completer_driver.dart
```diff
@@ -126,6 +126,7 @@ void dispatchNotification(String event, params) {
     _onServerStatus.add(true);
   }
 
+  // TODO: Should we be ignoring the ones that aren't marked 'isLast'?
```
lukechurch: Yes. This is because of the way in which the analysis server protocol works, it tries to stream completion results, the ones that aren't marked last are incomplete. Often very incomplete (aka empty).

For files the size of the ones we're looking at, we want to wait for exact completions (average time is only a few ms), so we only want the last results.

lukechurch:
lgtm, I'll replace the TODO comment

lib/src/completer_driver.dart
devoncarew: Ah, ok. Sounds like the each event contains the all the completions that are known at that point in time.

(I know we're just clients of the api, but this is a fairly pathological use of the stream api)

PR #88, opened 2015-03-22
Add support for counters

lukechurch:
This PR adds fast sharded counters for total aggregated usage of the various end points:

Currently it's tracking: Number of analyses, Compilations, Completions, Docs, number of lines analyzed and compiled. No PII is recorded only total usage numbers.

It also adds a counters query function to the API.

I'm going to push this to start being able to do in the wild performance testing on 1% of our queries. It also needs a bit more testing to verify that the implementation is properly concurrency safe.

FYI @devoncarew @clayberg

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -33,6 +33,11 @@ abstract class SourceRequestRecorder {
   Future record(String verb, String source, [int offset]);
 }
 
+abstract class PersistentCounter {
+  Future increment(String name, {int increment : 1});
+  Future<int> getTotal(String name);
+}
+
 class SourceRequest {
   @ApiProperty(required: true)
   String source;
@@ -43,6 +48,17 @@ class CompileResponse {
   final String result;
 
   CompileResponse(this.result);
+}
+
+class CounterRequest {
+  @ApiProperty(required: true)
+  String name;
+}
+
+class CounterResponse {
+  final int count;
+
+  CounterResponse(this.count);
 }
 
 class CompleteResponse {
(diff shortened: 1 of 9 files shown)


lib/src/completer_driver.dart
```diff
@@ -126,7 +126,8 @@ void dispatchNotification(String event, params) {
     _onServerStatus.add(true);
   }
 
-  // TODO: Should we be ignoring the ones that aren't marked 'isLast'?
+  // Ignore all but the last compeltion result. This means that we get a
```
devoncarew: sp: `compeltion`

lib/src/common_server.dart
```diff
@@ -149,17 +176,19 @@ class CommonServer {
...
     }
     Stopwatch watch = new Stopwatch()..start();
     srcRequestRecorder.record("ANALYZE", source);
     try {
-      return analyzer.analyze(source).then((AnalysisResults results) {
+      return analyzer.analyze(source).then((AnalysisResults results) async {
         int lineCount = source.split('\n').length;
         int ms = watch.elapsedMilliseconds;
         _logger.info('PERF: Analyzed ${lineCount} lines of Dart in ${ms}ms.');
+        await counter.increment("Analyses");
```
devoncarew: Perhaps instead of `await`, you just call the `increment()` method, and not wait for it's completion? I assume it'll be crazy fast, but we might want to return the analysis results w/o waiting for the counter to increment.

lib/src/common_server.dart
```diff
@@ -149,17 +176,19 @@ class CommonServer {
...
+        await counter.increment("Analyzed-Lines", increment: lineCount);
```
devoncarew: Also, perhaps track the number of analysis issues found? `results.issues.length`

devoncarew:
Great to have these metrics available!

lib/src/common_server.dart
lukechurch: Yes, I tried to do this. There is a complex bug, that these transaction backed datastore mutations don't seem to behave like normal datastore writes. If they aren't completed during the query they appear to cause the datastore context to get destroyed which results in them failing to apply. This doesn't happen with the normal writes on e.g. source recorder.

I need to talk to the GAE folks to understand what is going on here, in the meantime await avoids the problem.

I'll add a TODO to get this removed and file an issue.

PR #89, opened 2015-03-24
Move counter to background writes

lukechurch:
FYI @devoncarew The problem with the writes failing was in the use of the dbservices object, it's only available when in the context of a request. So if we want to do background writes, we need to keep this rather than re-querying for it.

Now we do this, we can move to background writes, which buys us a bit less latency - and more important - much lower variability as we're not awaiting a datastore RPC.

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,4 +1,4 @@
-version: cc-counter
+version: cc-bg-counter
 runtime: custom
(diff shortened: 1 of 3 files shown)


devoncarew:
Cool - good to have tracked this down.

PR #91, opened 2015-03-25
sends parameter info in a list for use in dartdoc

kasperpeulen:
for example, the dartdoc for Stream.listen would have now the extra key

```
  "parameters": [
   "(MouseEvent) → void onData",
   "{Function onError}",
   "{() → void onDone}",
   "{bool cancelOnError}"
  ]
  ```

I think this is useful for example, if you want to show parameter info and you want to do some formatting with that info.

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -137,6 +137,13 @@ class Analyzer {
         info['description'] = '${element}';
         info['kind'] = element.kind.displayName;
 
+        //parameters for functions and methods
+        if (element is ExecutableElement) {
+          List<String> list = [];
+          element.parameters.forEach((par) => list.add('${par}'));
+          info['parameters'] = list;
+        }
+
         // library
         LibraryElement library = element.library;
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/src/analyzer.dart [hunk 1]
```diff
@@ -137,6 +137,13 @@ class Analyzer {
         info['description'] = '${element}';
         info['kind'] = element.kind.displayName;
 
+        //parameters for functions and methods
+        if (element is ExecutableElement) {
+          List<String> list = [];
+          element.parameters.forEach((par) => list.add('${par}'));
```
devoncarew: The build is complaining about this line (https://travis-ci.org/dart-lang/dart-services/builds/55840151#L240). You'll probably need to add an explicit cast:
    ExecutableElement ee = element.
    ee.parameters...

lib/src/analyzer.dart
```diff
@@ -137,6 +137,14 @@ class Analyzer {
         info['description'] = '${element}';
         info['kind'] = element.kind.displayName;
 
+        //parameters for functions and methods
+        if (element is ExecutableElement) {
+          List<String> list = [];
+          ExecutableElement el = element;
```
lukechurch: This line isn't strictly necessary?

lukechurch:
Thanks, lgtm!

PR #96, opened 2015-03-25
Simplify error handling

lukechurch:
Thanks for the suggestion @devoncarew

### DIFF
--- a/lib/src/completer_driver.dart
+++ b/lib/src/completer_driver.dart
@@ -35,9 +35,6 @@ StreamController<bool> _onServerStatus;
 Stream<Map> completionResults;
 StreamController<Map> _onCompletionResults;
 
-Stream<Map> errors;
-StreamController<Map> _onErrors;
-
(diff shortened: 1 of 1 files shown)


lib/src/completer_driver.dart
```diff
@@ -100,17 +94,8 @@ Future<Map> _complete(String src, int offset) async {
...
-
-  MergeStream completionOrError = new MergeStream();
-  completionOrError.add(completionResults);
-
-  // We want to watch on the stream of errors to make sure that the request
-  // returns, even if it's going to fail.
-  completionOrError.add(errors);
-
-  return completionOrError.stream.first;
+  return completionResults.handleError(
```
devoncarew: not sure this line is necessary - any error on `completionResults` would bubble through w/o the throw added.

lukechurch:
Merging this for now given that this code is going to get cleaned up shortly anyhow

PR #97, opened 2015-03-26
refactor as integration

devoncarew:
Work in progress PR to refactor the analysis server integration. Goals:

- simplify the async code for returning completion results
- remove machine specific configuration settings
- add integration tests
- add benchmarks

(PR is currently not functional)


lukechurch:
@devoncarew thanks for this. Lets chat when you get in today.

I'd like to split this up into bits. Once we remove machine specific settings, we can fold in the integration tests and benchmarks.

We can then talk separately about what your thoughts about the async design change is - and I can explain what the current tangled mess is actually doing.

devoncarew:
closing as discussed

PR #104, opened 2015-03-29
Remove package root

lukechurch:
Remove the need to pass the package root to the analysis server

### DIFF
--- a/lib/src/completer_driver.dart
+++ b/lib/src/completer_driver.dart
@@ -19,8 +19,7 @@ io.Directory sourceDirectory = io.Directory.systemTemp.createTempSync('analysisS
 // GAE configurations
 // TODO(lukechurch): Migrate into a ctor
 String SDK = null;
-String PACKAGE_ROOT = '/app/packages';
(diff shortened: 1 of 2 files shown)


lib/src/completer_driver.dart
```diff
@@ -19,8 +19,7 @@ io.Directory sourceDirectory = io.Directory.systemTemp.createTempSync('analysisS
 // GAE configurations
 // TODO(lukechurch): Migrate into a ctor
 String SDK = null;
-String PACKAGE_ROOT = '/app/packages';
-String SERVER_PATH = "/app/lib/src/analysis_server_server.dart";
+String SERVER_PATH = "/app/bin/_analysis_server_entry.dart";
```
devoncarew: I suspect this will always be run from the project root, in which case you could use `bin/_analysis_server_entry.dart`.  That'll make it machine agnostic.

lukechurch: Yeah, that's what I'm working on right now :)

PR #107, opened 2015-03-31
Refactor Analysis Server integration

lukechurch:
@devoncarew This takes a number of steps along the route we discussed.

It: 
- adds a completion benchmark
- adds the first end-to-end simple test of completion
- Refactors the Analysis Server into a wrapped class rather than using top level methods

There's still more to go, but this feels like a significant enough step in the right direction.


benchmark/bench.dart
```diff
@@ -24,6 +26,10 @@ void main(List<String> args) {
     new AnalyzerBenchmark('hellohtml', sampleCodeWeb),
     new AnalyzerBenchmark('sunflower', _sunflower),
 
+    new AnalysisServerBenchmark('hello', sampleCode),
```
devoncarew: What about calling this class `CompletionBenchmark`? The other class names (not part of this PR) should be updated at some point.

benchmark/bench.dart
```diff
@@ -10,6 +10,8 @@ import 'package:services/src/analyzer.dart';
 import 'package:services/src/bench.dart';
 import 'package:services/src/common.dart';
 import 'package:services/src/compiler.dart';
+import 'package:services/src/analysis_server.dart';
```
devoncarew: nit: sorting

benchmark/bench.dart
devoncarew: Also, we run each benchmark over three different types of sources. In the future, we should just drop this down to one moderately complex source file.

lib/src/analysis_server.dart
```diff
@@ -1,182 +1,152 @@
...
-library services.completer_driver;
+/// A wrapper around an analysis server instance
+library services.analysis_server;
 
 import 'dart:io' as io;
 import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
 import 'package:logging/logging.dart';
+import 'dart:convert' show JSON;
```
devoncarew: nit: sorting - lexical and grouped by import type

lib/src/analysis_server.dart
```diff
@@ -1,182 +1,152 @@
...
+      return new CompleteResponse(
+          response['replacementOffset'], response['replacementLength'],
+          results);
+    });
+  }
 
-  _onCompletionResults = new StreamController(sync: true);
-  completionResults = _onCompletionResults.stream.asBroadcastStream();
+  /// Cleanly shutdown the Analysis Server.
+  Future shutdown() async => serverConnection.sendServerShutdown();
```
devoncarew: Is it necessary to use the async keyword here?

lib/src/analysis_server.dart
```diff
@@ -1,182 +1,152 @@
...
 
-  server.listenToOutput(dispatchNotification);
-  server.sendServerSetSubscriptions([ServerService.STATUS]);
+    return serverConnection.completionResults.handleError(
+        (error) => throw "Completion failed").first;
+  }
 
-  _logger.fine("Server Set Subscriptions completed");
+  /// Warm up the analysis server to be ready for use.
+  Future warmup([bool useHtml = false]) =>
```
devoncarew: possibly prefer a named optional param here

lib/src/analysis_server.dart
```diff
@@ -465,3 +471,34 @@ class Server {
   }
 }
 
+class CompleteResponse {
```
devoncarew: `CompleteResponse` or `CompletionResponse`?

devoncarew: Also, I think all the rpc response classes should go into the common_server.dart file.

devoncarew: Ah, it was CompleteResonse before.

test/analysis_server_test.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2015, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library services.analyzer_server_test;
+
+import 'package:services/src/analysis_server.dart';
+import 'package:grinder/grinder.dart' as grinder;
```
devoncarew: You could use the `cli_utils` package here instead.

test/analysis_server_test.dart
```diff
@@ -0,0 +1,48 @@
...
+  group('analysis_server', () {
+    setUp(() {
+      analysisServer = new AnalysisServerWrapper(sdkPath);
+    });
+
+    tearDown(() {
+      analysisServer.shutdown();
+    });
+
+    test('simple', () {
```
devoncarew: woot!

devoncarew:
lgtm! I saw a build failure - I re-started the bot as it wasn't clear what happened. I _think_ all the tests ran, but then it hung when starting to run the benchmarks.

test/common_server_test.dart
```diff
@@ -84,11 +84,12 @@ void defineTests() {
           contains('failed with errors: [error, line 2] Expected'));
     });
 
-    /* XXX: Disabled below tests as they are causing the dart vm to crash.
     test('complete', () async {
       var json = {'source': 'void main() {print("foo");}', 'offset': 1};
       var response = await _sendPostRequest('dartservices/v1/complete', json);
-      expect(response.status, 501);
+      expect(response.status, 200);
```
devoncarew: You could try commenting these three tests back out. Perhaps the AS is not being terminated, and the VM is just never exiting (waiting on an open socket)?

benchmark/bench.dart
lukechurch: sg. Will address in follow on PR.

benchmark/bench.dart
lukechurch: done

lib/src/analysis_server.dart
lukechurch: Hmmm, there really shouldn't be two separate imports for this. Fixed.

lib/src/analysis_server.dart
lukechurch: Nope, legacy. Thanks, removed.

lib/src/analysis_server.dart
lukechurch: This is consistent with the other APIs, however, I agree. I'll adjust them all in one go.

test/analysis_server_test.dart
lukechurch: done. I've adjusted the other test targets that used grinder as well.

test/common_server_test.dart
lukechurch: Yeah, I think you're probably right. Done.

lukechurch: There was also a problem with the benchmarks' method being called 'teardown' not 'tearDown'.

PR #108, opened 2015-04-02
show parent information in dartdoc

kasperpeulen:
two examples

```json
{
   "info":{
      "name":"isEven",
      "description":"get isEven → bool",
      "kind":"getter",
      "parent":"int",
      "parameters":[

      ],
      "libraryName":"dart.core",
      "dartdoc":"Returns true if and only if this integer is even.",
      "staticType":"bool"
   }
}
```

```json
{
   "info":{
      "name":"main",
      "description":"main() → void",
      "kind":"function",
      "parent":"main.dart",
      "parameters":[

      ],
      "staticType":"() → void"
   }
}
```

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -133,9 +133,9 @@ class Analyzer {
 
         // Name and description.
         if (element.name != null) info['name'] = element.name;
-        //if (element.displayName != null) info['displayName'] = element.displayName;
         info['description'] = '${element}';
         info['kind'] = element.kind.displayName;
+        info['parent'] = '${element.enclosingElement}';
 
         //parameters for functions and methods
         if (element is ExecutableElement) {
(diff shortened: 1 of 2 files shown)


lib/src/analyzer.dart
```diff
@@ -133,9 +133,9 @@ class Analyzer {
 
         // Name and description.
         if (element.name != null) info['name'] = element.name;
-        //if (element.displayName != null) info['displayName'] = element.displayName;
         info['description'] = '${element}';
         info['kind'] = element.kind.displayName;
+        info['parent'] = '${element.enclosingElement}';
```
lukechurch: parent has a lot of meanings. Can we call this 'enclosingElement'?

lukechurch:
1 rename, and then I think we're good to go.

lukechurch:
Actually I think I'd rather than a more semantic form for 'enclosingElement' - so that you can tell if it's going to give you.

lukechurch:
Now serving from test deployment doc-test.....

PR #109, opened 2015-04-02
Return DOM Name to support directing the user to better documentation

kasperpeulen:


lib/src/analyzer.dart
```diff
@@ -163,6 +163,17 @@ class Analyzer {
               info['libraryName'] = library.name;
             }
           }
+          if (library.location.toString() == "dart:html") {
+            if (element.metadata != null) {
+              element.metadata.forEach((ElementAnnotationImpl e) {
+                if (e.toString().startsWith("@DomName")) {
+                  if (e.evaluationResult.value.fields["name"] != null) {
```
lukechurch: Lets reduce the redundant lookup

var nameValue = e.evaluationResult.value.fields["name"];
if (nameValue != null) info["DomName"] = nameValue.value;

lib/src/analyzer.dart
```diff
@@ -163,6 +163,17 @@ class Analyzer {
               info['libraryName'] = library.name;
             }
           }
+          if (library.location.toString() == "dart:html") {
+            if (element.metadata != null) {
+              element.metadata.forEach((ElementAnnotationImpl e) {
```
lukechurch: It's better to do a 
for (ElementAnnotationImpl in element.metadata) here

As it means that we can add in a break; statement after we have found the DomName attribution, which short-cuts the rest of the loop.

lukechurch:
e.evaluationResult is strictly bound to constant evaluation, so the side effects don't pose a risk.

lukechurch:
@kasperpeulen Thanks.

I'm going to merge this in and run it through stress testing.e

lib/src/analyzer.dart
```diff
@@ -20,6 +20,7 @@ import 'package:analyzer/src/generated/source_io.dart';
 import 'package:logging/logging.dart';
 
 import 'common.dart';
+import 'package:analyzer/src/generated/constant.dart';
```
devoncarew: nit: import ordering

Issue #106, opened 2015-03-30
Completion order on user defined classes is incorrect


devoncarew:
This ordering looks odd to me:

![screen shot 2015-03-30 at 12 00 18 pm](https://cloud.githubusercontent.com/assets/1269969/6904280/9a5b159a-d6d4-11e4-916d-15b0b3706b5b.png)

From this gist: dartpad.dartlang.org/7eb492810a52928c8e0e

I would expect lexical sorting - to help users find items - unless there were strong signals to move particular items higher in the list.

lukechurch:
Yup, that's a bug.

devoncarew:
It's in that linked gist.

lukechurch:
Sorry, I didn't see it. thanks

PR #111, opened 2015-04-03
Fall back to lexical sorting

lukechurch:
If there is no obvious relevance sorting, fall back to lexical

This fixes: #106 It's not a complete solution, but should fix a lot of the silliest problems.

FYI @devoncarew


devoncarew:
lgtm

PR #115, opened 2015-04-04
Support quick fixes in the API

lukechurch:
This PR adds support for quick fixes, it extends the test tool and adds a minimal test case.

It also addresses some refactoring consolidating the API classes into a separate file.

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -55,6 +55,14 @@ class AnalysisServerWrapper {
     });
   }
 
+  Future<FixesResponse> getFixes(String src, int offset) async {
+    var results = _getFixesImpl(src, offset);
+
+    return results.then((fixes) {
+        return new FixesResponse(fixes.fixes);
+    });
+  }
+
   /// Cleanly shutdown the Analysis Server.
   Future shutdown() => serverConnection.sendServerShutdown();
 
@@ -68,6 +76,13 @@ class AnalysisServerWrapper {
 
     return serverConnection.completionResults.handleError(
         (error) => throw "Completion failed").first;
+  }
+
+  _getFixesImpl(String src, int offset) async {
+    await serverConnection._ensureSetup();
+
+    serverConnection.sendAddOverlay(src);
+    return await serverConnection.sendGetFixes(offset);
   }
 
   /// Warm up the analysis server to be ready for use.
@@ -405,6 +420,14 @@ class _Server {
       });
     }
 
+    Future<EditGetFixesResult> sendGetFixes(int offset) {
+      String file = psuedoFilePath;
+      var params = new EditGetFixesParams(file, offset).toJson();
+      return send("edit.getFixes", params).then((result) {
+        ResponseDecoder decoder = new ResponseDecoder(null);
+        return new EditGetFixesResult.fromJson(decoder, 'result', result);
+      });
+    }
 
     Future<AnalysisUpdateContentResult> sendAddOverlay(String contents) {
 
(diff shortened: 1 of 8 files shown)


lukechurch:
I'd like to bring this in to do some combined longer running tests. TBR @devoncarew

kasperpeulen:
+1 awesome :) can I use this already ?

lib/src/analysis_server.dart
```diff
@@ -70,6 +78,13 @@ class AnalysisServerWrapper {
         (error) => throw "Completion failed").first;
   }
 
+  _getFixesImpl(String src, int offset) async {
```
devoncarew: Can you add a return type?

lib/src/api_classes.dart
```diff
@@ -0,0 +1,173 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
devoncarew: nit: 2015

lib/src/api_classes.dart
```diff
@@ -0,0 +1,173 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+library services.api_classes;
```
devoncarew: `/// All classes exported over the RPC protocol.`

lib/src/api_classes.dart
```diff
@@ -0,0 +1,173 @@
...
+  }
+}
+
+class FixesResponse {
+  final List<ProblemFix> fixes;
+
+  FixesResponse(List<AnalysisErrorFixes> analysisErrorFixes) :
+    this.fixes = _convert(analysisErrorFixes);
+
+  /**
```
devoncarew: I don't think this dartdoc comment is comment for this method.

lib/src/api_classes.dart
```diff
@@ -0,0 +1,173 @@
...
+   * Convert any non-string values from the contained maps.
+   */
+  static List<ProblemFix> _convert(List<AnalysisErrorFixes> list) {
+    var problemsAndFixes = new List<ProblemFix>();
+    list.forEach((analysisErrorFix)
```
devoncarew: `analysisErrorFix` ==> `fix` for brevity?

lib/src/api_classes.dart
```diff
@@ -0,0 +1,173 @@
...
+        => problemsAndFixes.add(_convertAnalysisErrorFix(analysisErrorFix)));
+
+    return problemsAndFixes;
+  }
+
+  static ProblemFix _convertAnalysisErrorFix(AnalysisErrorFixes analysisErrFix) {
```
devoncarew: `analysisErrFix` => `fixes`?

lib/src/api_classes.dart
```diff
@@ -0,0 +1,173 @@
...
+    String problemMessage = analysisErrFix.error.message;
+    int problemOffset = analysisErrFix.error.location.offset;
+    int problemLength = analysisErrFix.error.location.length;
+
+    List<Fix> possibleFixes = new List<Fix>();
+
+    for (var sourceChange in analysisErrFix.fixes) {
+      List<Edit> edits = new List<Edit>();
+      for (var sourceFileEdit in sourceChange.edits) {
+        for (var sourceEdit in sourceFileEdit.edits) {
```
devoncarew: Should we verify that all the fixes are trying to modify the source in our single source file? Is there any possibility that they would be trying to modify other files? They're all targeted at our single `Source`?

lib/src/analysis_server.dart
lukechurch: Will do.

lib/src/api_classes.dart
lukechurch: +1

lib/src/api_classes.dart
lukechurch: Thanks, it's legacy.

lib/src/api_classes.dart
lukechurch: +1

lib/src/api_classes.dart
lukechurch: I don't think there is, but adding in a test to ensure that that is always true seems like a good idea. I'll do that

lib/src/api_classes.dart
lukechurch: I'd rather we put something that made it very explicit that these are from analysis server otherwise this code gets confusing. I'll go for analysisFixes for now?

devoncarew: Sure, sgtm.

PR #117, opened 2015-04-04
Update support v1 client

lukechurch:
@kasperpeulen This updates the JSON and v1 client to support completion requests

@devoncarew The commands send to the discovery tool seems to have changed - from generate -> package. Can you confirm?

### DIFF
--- a/doc/generated/v1.dart
+++ b/doc/generated/v1.dart
@@ -1,3 +1,5 @@
+// This is a generated file (see the discoveryapis_generator project).
+
 library dartservices_clientlib.dartservices.v1;
 
 import 'dart:core' as core;
@@ -14,7 +16,6 @@ export 'package:_discoveryapis_commons/_discoveryapis_commons.dart' show
 
 const core.String USER_AGENT = 'dart-api-client dartservices/v1';
 
-/** Not documented yet. */
 class DartservicesApi {
 
   final commons.ApiRequester _requester;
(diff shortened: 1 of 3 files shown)


doc/generated/v1.dart
```diff
@@ -603,12 +650,145 @@ class DocumentResponse {
 }
 
 
-/** Not documented yet. */
+class Edit {
```
devoncarew: This is a very short, fairly generic class name. Perhaps consider renaming the source class this is generated from (`FixEdit`? `SourceEdit`?)

doc/generated/v1.dart
```diff
@@ -603,12 +650,145 @@ class DocumentResponse {
...
+    var _json = new core.Map();
+    if (fixes != null) {
+      _json["fixes"] = fixes.map((value) => (value).toJson()).toList();
+    }
+    return _json;
+  }
+}
+
+
+class ProblemFix {
```
devoncarew: Some docs on the source class would help here, I'm not 100% what a problem fix is.

tool/grind.dart
```diff
@@ -39,7 +39,7 @@ discovery(GrinderContext context) {
   // Generate the Dart library from the json discovery file.
   Pub.global.activate(context, 'discoveryapis_generator');
   Pub.global.run(context, 'discoveryapis_generator:generate', arguments: [
-    'generate',
+    'package',
```
devoncarew: Yup, @lukechurch, this did change. I didn't realize that they had published the new version. Also, slightly prettier output :)

lukechurch:
Yeah, these things do end up being somewhat torn out of context.

I'll rename to more specific and document.

devoncarew:
lgtm! Yup, the generator's command changed. It can also generate stand-alone libraries now.

PR #118, opened 2015-04-04
Address review comments

lukechurch:
Thanks @devoncarew


lib/src/api_classes.dart
```diff
@@ -108,26 +109,38 @@ class FixesResponse {
...
+
+      // A fix that tries to modify other files is considered invalid.
+
+      bool invalidFix = false;
       for (var sourceFileEdit in sourceChange.edits) {
+
+        // TODO(lukechurch): replace this with a more reliable test based
+        // on the psuedo file name in Analysis Server
+        if (!sourceFileEdit.file.endsWith("/main.dart")) {
+          invalidFix = true;
```
devoncarew: We should also log invalid ones, just so we known to try and track down what those fixes were for.

PR #119, opened 2015-04-04
issue has fixes (DNM)

kasperpeulen:
just an idea if we could maybe check if in issue has fixes in this way


lukechurch:
As discussed, I'd like us to only use one of the two technologies we have for each of these {Analysis Engine, Analyzer}. For now, lets do the querying in the UI. Thanks

PR #121, opened 2015-04-06
Renames to API classes

lukechurch:
@devoncarew PTAL

Do you find these any more explanatory?

Cheers

### DIFF
--- a/doc/generated/v1.dart
+++ b/doc/generated/v1.dart
@@ -465,6 +465,8 @@ class AnalysisIssue {
 
   core.int charStart;
 
+  core.bool hasFix;
+
   core.String kind;
 
   core.int line;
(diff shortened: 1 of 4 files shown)


doc/generated/v1.dart
```diff
@@ -544,6 +544,36 @@ class AnalysisResults {
 }
 
 
+class CandidateFix {
```
devoncarew: This being just `Fix` made sense to me (and I didn't think was likely to conflict with other class names); whichever seems best to you.

doc/generated/v1.dart
```diff
@@ -650,83 +680,52 @@ class DocumentResponse {
...
-
-  Fix.fromJson(core.Map _json) {
-    if (_json.containsKey("edits")) {
-      edits = _json["edits"].map((value) => new Edit.fromJson(value)).toList();
-    }
-    if (_json.containsKey("message")) {
-      message = _json["message"];
-    }
-  }
+class ProblemAndFix {
```
devoncarew: `ProblemAndFixes`?

devoncarew:
Thanks for the renames! `Edit` definitely seemed a bit generic.

lukechurch:
@devoncarew Thanks, adjusted as discussed
@kasperpeulen I've put in a flag into the Analysis Result "HasFix". This can be used for setting the wrench icon.

PR #123, opened 2015-04-06
Enable has fixes for some specific cases

lukechurch:
@devoncarew PTAL

PR #124, opened 2015-04-06
make hasFixes default to false

kasperpeulen:


lukechurch:
@kasperpeulen Thanks. As you can see this causes the tests to regress. I've included it in #123 which should fix the problem both in the default and in the code that assembles the results before return

kasperpeulen:
nice :)

PR #125, opened 2015-04-07
extend dartdoc with information about the argument

### LINKED ISSUE #102: Extend dartdocinfo with information about the element as argument of a method/function
I'm thinking about extending the dartdocinfo so that it not only gives information about the offset as an expression, but also about information at the offset as an argument of a method or a function.

So for example, if take have:
```dart
void main() {
  var string = "test";
  string.substring(1, string.length<cursor>);
}
```

Currently dartdoc gives the following information:

```json
{
   "info":{
      "name":"length",
      "description":"get length → int",
      "kind":"getter",
      "parameters":[

      ],
      "libraryName":"dart.core",
      "dartdoc":"The length of the string.\n\nReturns the number of UTF-16 code units in this string. The number\nof [runes] might be fewer, if the string contains characters outside\nthe Basic Multilingual Plane (plane 0):\n\n    'Dart'.length;          // 4\n    'Dart'.runes.length;    // 4\n\n    var clef = '\\u{1D11E}';\n    clef.length;            // 2\n    clef.runes.length;      // 1",
      "staticType":"dynamic",
      "propagatedType":"int"
   }
}
```

I thought about extending it in this way:

```json
{
   "info":{
      "name":"length",
      "description":"get length → int",
      "kind":"getter",
      "parameters":[

      ],
      "libraryName":"dart.core",
      "dartdoc":"The length of the string.\n\nReturns the number of UTF-16 code units in this string. The number\nof [runes] might be fewer, if the string contains characters outside\nthe Basic Multilingual Plane (plane 0):\n\n    'Dart'.length;          // 4\n    'Dart'.runes.length;    // 4\n\n    var clef = '\\u{1D11E}';\n    clef.length;            // 2\n    clef.runes.length;      // 1",
      "staticType":"dynamic",
      "propagatedType":"int",
      "infoAsArgument":{
         "argumentOffset":[
            59,
            72
         ],
         "activeArgumentIndex":1,
         "lParenOffset":55,
         "rParenOffset":72,
         "parentInfo":{
            "name":"substring",
            "description":"String.substring(int startIndex, [int endIndex]) → String",
            "kind":"method",
            "parameters":[
               "int startIndex",
               "[int endIndex]"
            ],
            "libraryName":"dart.core",
            "dartdoc":"Returns the substring of this string that extends from [startIndex],\ninclusive, to [endIndex], exclusive.\n\n    var string = 'dartlang';\n    string.substring(1);    // 'artlang'\n    string.substring(1, 4); // 'art'",
            "staticType":"dynamic",
            "propagatedType":"String"
         }
      }
   }
}
```

kasperpeulen:
fix #102 

For example, if you have as input (offset:90):
```dart
import "dart:html";

void main() {
  new HttpRequest().request("url", mimeType:"test");
}
```

You get the following output:
```json
{
   "info":{
      "staticType":"String",
      "infoAsArgument":{
         "name":"mimeType: \"test\"",
         "kind":"NAMED",
         "indexActiveArgument":4,
         "parameters":[
            "String url",
            "{String method}",
            "{bool withCredentials}",
            "{String responseType}",
            "{String mimeType}",
            "{Map<String, String> requestHeaders}",
            "{dynamic sendData}",
            "{(ProgressEvent) → void onProgress}"
         ],
         "lParenOffset":62,
         "rParenOffset":85
      }
   }
}
```

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -109,6 +109,8 @@ class Analyzer {
 
   Map<String, String> _computeDartdocInfo(CompilationUnit unit, int offset) {
     AstNode node = new NodeLocator.con1(offset).searchWithin(unit);
+    Map info = {};
+    AstNode argument = node;
 
     if (node.parent is TypeName &&
         node.parent.parent is ConstructorName &&
@@ -123,7 +125,6 @@ class Analyzer {
 
     if (node is Expression) {
       Expression expression = node;
-      Map info = {};
 
       // element
       Element element = ElementLocator.locateWithOffset(expression, offset);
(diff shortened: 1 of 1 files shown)


lib/src/analyzer.dart
```diff
@@ -197,11 +198,92 @@ class Analyzer {
...
+        method = null;
+        break;
+      } else if (argument.parent is ArgumentList) {
+        method = argument.parent.parent;
+        break;
+      } else if (argument is ArgumentList) {
+        emptyArguments = true;
+        method = argument.parent;
+        break;
+      }
```
lukechurch: please add a cycle protection.

if (argument == argument.parent) throw 'cyclic AST';

kasperpeulen: I was not aware that there exist something like a self-cyclic AST, but I've added this now

lukechurch: There shouldn't be, but programming on servers needs to be defensive.
I.e. you try to assume where reasonably possible that the libraries you're consuming have bugs and to protect yourself.

On the server, unbound loops are esp. scary. Hence when you read a while loop you want to be able to prove to yourself that it will always exit :)

lib/src/analyzer.dart
```diff
@@ -197,11 +198,94 @@ class Analyzer {
...
+    if (method == null
+        || !(method is MethodInvocation || method is InstanceCreationExpression))
+      return null;
+
+    // method is either InstanceCreationExpression or MethodInvocation
+    // both have the getter argumentList
+    ArgumentList argList = method.argumentList;
+
+    var e = ElementLocator.locateWithOffset(method, method.offset);
+    if (!(e is ExecutableElement)) {
```
lukechurch: is!

lib/src/analyzer.dart
```diff
@@ -197,11 +198,94 @@ class Analyzer {
...
+      index = 0;
+    } else if (argument is NamedExpression) {
+      NamedExpression namedArgument = argument;
+      String nameOfArgument = namedArgument.name.label.toString();
+      //check if we can find the namedArgument in the list of parameters
+      try {
+        var activeParameter = parameters.firstWhere(
+                (parameter) => parameter.name == nameOfArgument);
+        index = parameters.indexOf(activeParameter);
+      } on StateError {
```
lukechurch: What would cause this?

lukechurch:
@kasperpeulen Sorry for the delay here - if you could address the comments and add a couple of unit tests, I think we're probably good to go here. Thanks!

lukechurch:
I'll pick up ownership of this branch and landing it

PR #128, opened 2015-04-08
Fix memory leak

lukechurch:
TBR @devoncarew 

This is backed from the 1.9.1 version. I'm going to push this and dogfood for a bit. I would like to have this in the prod server ASAP.


devoncarew:
lgtm

PR #136, opened 2015-04-13
plumb through package: support for the compile service call

devoncarew:
@lukechurch, additional work towards `package:` support:

- added a `enablePackages` compile time flag; currently off
- added a method to scan source code and extract package: imports
- added tests for the above
- added a helper class to parse source for imports, and provide methods to locate package resources
- plumbed all this through the `compile/` service call. analyzer and AS related calls not yet supported -

### DIFF
--- a/doc/generated/v1.dart
+++ b/doc/generated/v1.dart
@@ -3,10 +3,12 @@
 library dartservices_clientlib.dartservices.v1;
 
 import 'dart:core' as core;
+import 'dart:collection' as collection;
 import 'dart:async' as async;
 import 'dart:convert' as convert;
 
 import 'package:_discoveryapis_commons/_discoveryapis_commons.dart' as commons;
+import 'package:crypto/crypto.dart' as crypto;
 import 'package:http/http.dart' as http;
 
 export 'package:_discoveryapis_commons/_discoveryapis_commons.dart' show
--- a/doc/generated/v1.dart
+++ b/doc/generated/v1.dart
@@ -3,12 +3,10 @@
 library dartservices_clientlib.dartservices.v1;
 
 import 'dart:core' as core;
-import 'dart:collection' as collection;
 import 'dart:async' as async;
 import 'dart:convert' as convert;
 
 import 'package:_discoveryapis_commons/_discoveryapis_commons.dart' as commons;
-import 'package:crypto/crypto.dart' as crypto;
 import 'package:http/http.dart' as http;
 
 export 'package:_discoveryapis_commons/_discoveryapis_commons.dart' show
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -178,7 +178,9 @@ class CommonServer {
 
     // TODO(lukechurch): Remove this hack after
     // https://github.com/dart-lang/rpc/issues/15 lands
-    bool suppressCache = source.trim().endsWith("/** Suppress-Memcache **/");
+    var trimSrc = source.trim();
+    bool suppressCache = trimSrc.endsWith("/** Supress-Memcache **/") ||
+        trimSrc.endsWith("/** Suppress-Memcache **/");
 
     return checkCache("%%COMPILE:$sourceHash").then((String result) {
       if (!suppressCache && result != null) {
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -14,9 +14,13 @@ import 'api_classes.dart';
 import 'analysis_server.dart';
 import 'analyzer.dart';
 import 'compiler.dart';
+import 'pub.dart';
 
 final Duration _standardExpiration = new Duration(hours: 1);
 final Logger _logger = new Logger('common_server');
+
+/// Toggle to on to enable `package:` support.
+final bool enablePackages = false;
 
 abstract class ServerCache {
   Future<String> get(String key);
@@ -43,6 +47,7 @@ class CommonServer {
   final SourceRequestRecorder srcRequestRecorder;
   final PersistentCounter counter;
 
+  Pub pub;
   Analyzer analyzer;
   Compiler compiler;
   AnalysisServerWrapper analysisServer;
@@ -53,8 +58,11 @@ class CommonServer {
       this.counter) {
     hierarchicalLoggingEnabled = true;
     _logger.level = Level.ALL;
+
+    pub = enablePackages ? new Pub() : new Pub.mock();
+
     analyzer = new Analyzer(sdkPath);
-    compiler = new Compiler(sdkPath);
+    compiler = new Compiler(sdkPath, pub);
     analysisServer = new AnalysisServerWrapper(sdkPath);
   }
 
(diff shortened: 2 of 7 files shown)


doc/generated/v1.dart
```diff
@@ -3,12 +3,10 @@
 library dartservices_clientlib.dartservices.v1;
 
 import 'dart:core' as core;
-import 'dart:collection' as collection;
```
lukechurch: Are these removed from the auto-genned version library? If not, lets leave them in.

lib/src/common_server.dart
```diff
@@ -14,10 +14,14 @@ import 'api_classes.dart';
 import 'analysis_server.dart';
 import 'analyzer.dart';
 import 'compiler.dart';
+import 'pub.dart';
 
 final Duration _standardExpiration = new Duration(hours: 1);
 final Logger _logger = new Logger('common_server');
 
+/// Toggle to on to enable `package:` support.
+final bool enablePackages = false;
```
lukechurch: ENABLE_PACKAGES?

lib/src/common_server.dart
```diff
@@ -170,14 +178,14 @@ class CommonServer {
 
     // TODO(lukechurch): Remove this hack after
     // https://github.com/dart-lang/rpc/issues/15 lands
-    bool supressCache = source.trim().endsWith("/** Supress-Memcache **/");
+    bool suppressCache = source.trim().endsWith("/** Suppress-Memcache **/");
```
lukechurch: Lets not do this in this PR - as it needs to be synced with the uptime services

lukechurch: I'll follow up fixing this.

lib/src/compiler.dart
```diff
@@ -27,30 +28,37 @@ Logger _logger = new Logger('compiler');
...
-        provider.getInitialUri(),
-        new Uri(scheme: 'sdk', path: '/'),
-        new Uri(scheme: 'package', path: '/'),
-        provider.inputProvider,
-        result._diagnosticHandler,
-        ['--no-source-maps'],
-        result._outputProvider).then((_) {
-      result._problems.sort();
-      return result;
+    Future<PubHelper> f = pub == null ?
```
lukechurch: Consider a more expressive var name?

lukechurch: Is this really just because it's a place holder for future?

lib/src/compiler.dart
```diff
@@ -27,30 +28,37 @@ Logger _logger = new Logger('compiler');
...
+        new Future.value() : pub.createPubHelperForSource(input);
+
+    return f.then((pubHelper) {
+      _CompilerProvider provider = new _CompilerProvider(_sdk, input, pubHelper);
```
lukechurch: We can use async/await style in the server. It's used pretty extensively.

I think this would turn into var pubHelper = await ...

lukechurch:
Will review the parser parts of this code in a couple of hours

doc/generated/v1.dart
devoncarew: I filed an issue against the generation library - they should not be adding these. It means that our bot has to be lenient towards analysis warnings.

Will add them back in...

lib/src/common_server.dart
devoncarew: I think the style guide has changed to prefer lowerCaseConstants.

lib/src/common_server.dart
devoncarew: Oops, I should have caught that.

lib/src/compiler.dart
devoncarew: Will improve the name; if is either a no-op or an actual call to the pub service.

lib/src/compiler.dart
devoncarew: switched to async await ...

devoncarew:
Updated for the current comments -

lib/src/pub.dart
```diff
@@ -186,6 +203,82 @@ class Pub {
...
+ */
+class PubHelper {
+  final Pub pub;
+  final List<PackageInfo> packages;
+
+  PubHelper._(this.pub, this.packages);
+
+  bool get hasPackages => packages.isNotEmpty;
+
+  Future<String> getPackageContentsAsync(String pathFragment) {
```
lukechurch: If I read this method correctly, it only supports a subset of possible import semantics? We should probably at least list those in a comment?

lib/src/pub.dart
```diff
@@ -186,6 +203,82 @@ class Pub {
...
+    String packageName = pathFragment.substring(0, index);
+    pathFragment = pathFragment.substring(index + 1);
+
+    PackageInfo package = getPackage(packageName);
+    if (package == null) {
+      return new Future.error('package not found: ${packageName}');
+    }
+
+    return pub.getPackageLibDir(package).then((dir) {
+      File file = new File(path.join(dir.path, pathFragment));
```
lukechurch: I'd like this access to be recorded in the log

lib/src/pub.dart
```diff
@@ -213,3 +306,82 @@ class PackageInfo {
 
   String toString() => '[${name}: ${version}]';
 }
+
+Set<String> getAllImportsFor(String dartSource) {
+  if (dartSource == null) return new Set();
+
+  Scanner scanner = new Scanner(
+      new StringSource(dartSource, 'temp.dart'),
```
lukechurch: Is it important that elsewhere we've called this main.dart?

lib/src/pub.dart
```diff
@@ -213,3 +306,82 @@ class PackageInfo {
 
   String toString() => '[${name}: ${version}]';
 }
+
+Set<String> getAllImportsFor(String dartSource) {
```
lukechurch: Lets be explicit about where it is that we're cleaning this up.

consider calling this "getAllUnsafeImportsFor"

lib/src/pub.dart
```diff
@@ -213,3 +306,82 @@ class PackageInfo {
...
+  }
+
+  return imports;
+}
+
+/**
+ * Return the list of packages that are imported from the given imports. These
+ * packages are sanitized defensively.
+ */
+Set<String> filterPackagesFromImports(Set<String> allImports) {
```
lukechurch: consider: filterSafePackagesFromImports

lib/src/pub.dart
lukechurch: Maybe we could add asserts to the front of this method?

test/pub_test.dart
```diff
@@ -78,6 +78,88 @@ void defineTests() {
...
+void main() { }
+''';
+      expect(getAllImportsFor(source), unorderedEquals(['dart:math', 'package:foo/foo.dart']));
+    });
+
+    test('two', () {
+      final String source = '''
+library woot;
+import 'dart:math';
+ import 'package:foo/foo.dart';
```
lukechurch: ws deliberate?

test/pub_test.dart
```diff
@@ -78,6 +78,88 @@ void defineTests() {
...
+      final String source = '''
+import 'package:foo/foo.dart';
+import 'package:bar/bar.dart';
+void main() { }
+''';
+      expect(filterPackagesFromImports(getAllImportsFor(source)),
+          unorderedEquals(['foo', 'bar']));
+    });
+
+    test('defensive', () {
```
lukechurch: Could you also add a neg test for 

import 'dart:../bar.dart'?

and 

import '../foo.dart'

lukechurch:
Thanks @devoncarew once comments are addressed, especially 

- a comment clarifying what 'getPackageContentsAsync' is expecting
- package file access being logged
- A rename annotating where the safety boundary is on the methods
- A couple more negative tests

This LGTM.

lib/src/pub.dart
devoncarew: It'll support all package: imports; I'll clarify that in the dartdoc.

lib/src/pub.dart
devoncarew: done

lib/src/pub.dart
devoncarew: will change

lib/src/pub.dart
devoncarew: done

lib/src/pub.dart
devoncarew: done

test/pub_test.dart
devoncarew: yup, to exercise the parser

test/pub_test.dart
devoncarew: added

devoncarew:
merging -

PR #141, opened 2015-04-16
Support format API

lukechurch:
This enables support for formatting.

It's not ready to be deployed until a solution to https://code.google.com/p/dart/issues/detail?id=23230 is agreed.

This is more to give @kasperpeulen a sense of what the API is going to be like. Note it formats the whole file at once. The region selection in the Server API doesn't actually allow selective formatting, it just allows the selection to be persisted between formats.

Note that guarding the API with an analysis call is needed due to the way the format API currently works.

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -23,10 +23,10 @@ typedef void NotificationProcessor(String event, params);
 final Logger _logger = new Logger('analysis_server');
 
 /**
- * Compile time flag to determine wheter we should dump
- * the communication with the server to stdout.
+ * Flag to determine wheter we should dump the communication
+ * with the server to stdout.
  */
-final bool _DUMP_SERVER_MESSAGES = false;
+bool dumpServerMessages = false;
 
 final _WARMUP_SRC_HTML = "import 'dart:html'; main() { int b = 2;  b++;   b. }";
 final _WARMUP_SRC = "main() { int b = 2;  b++;   b. }";
@@ -95,10 +95,12 @@ class AnalysisServerWrapper {
     await serverConnection.sendAddOverlay(src);
     await serverConnection.analysisComplete.first;
     await serverConnection.sendCompletionGetSuggestions(offset);
-    serverConnection.sendRemoveOverlay();
 
     return serverConnection.completionResults.handleError(
-        (error) => throw "Completion failed").first;
+        (error) => throw "Completion failed").first.then((ret) async {
+          await serverConnection.sendRemoveOverlay();
+          return ret;
+    });
   }
 
   Future<EditGetFixesResult> _getFixesImpl(String src, int offset) async {
@@ -107,7 +109,7 @@ class AnalysisServerWrapper {
     await serverConnection.sendAddOverlay(src);
     await serverConnection.analysisComplete.first;
     var fixes = await serverConnection.sendGetFixes(offset);
-    serverConnection.sendRemoveOverlay();
+    await serverConnection.sendRemoveOverlay();
     return fixes;
   }
 
@@ -116,7 +118,7 @@ class AnalysisServerWrapper {
     await serverConnection.sendAddOverlay(src);
     await serverConnection.analysisComplete.first;
     var formatResult = await serverConnection.sendFormat(offset);
-    serverConnection.sendRemoveOverlay();
+    await serverConnection.sendRemoveOverlay();
     return formatResult;
   }
 
@@ -200,7 +202,6 @@ class _Server {
     isSetup = true;
 
     _logger.fine("Setup done");
-
     return analysisComplete.first;
   }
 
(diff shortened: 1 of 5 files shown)


lukechurch:
I think the build failure is caused by a flaky pub test.

devoncarew:
The test failure should be fixed in master.

lib/src/api_classes.dart
```diff
@@ -186,6 +186,23 @@ class CandidateFix {
   CandidateFix(this.message, this.edits);
 }
 
+/*
+ * Represents a reformatting of the code.
+ */
+class FormatResponse {
+  List<SourceEdit> edits;
```
devoncarew: To make this easier for clients, could we just return the new, formatted, source? And perhaps an optional cursor position.

lib/src/common_server.dart
```diff
@@ -247,6 +261,12 @@ class CommonServer {
       return analysisServer.getFixes(source, offset);
     }
 
+  Future<FormatResponse> _format(String source) async {
```
devoncarew: This should take an optional int for the current cursor pos.

lib/src/analysis_server.dart
```diff
@@ -35,14 +36,20 @@ final _SERVER_PATH = "bin/_analysis_server_entry.dart";
...
+  // TODO(lukechurch): Replace this with listening to server errors and
+  // warnings.
+  Analyzer analyzer;
+
   /// Instance to handle communication with the server.
   _Server serverConnection;
 
   AnalysisServerWrapper(this.sdkPath) {
     serverConnection = new _Server(this.sdkPath);
+    analyzer = new Analyzer(this.sdkPath);
```
devoncarew: `Analyzer` instances are heavy-weight objects; they use significant amounts of memory. We should re-use the one from the common_server class. In general, I think:

- the AS's `edit.format` format call should be resilient to code with errors. It should either refuse to format it, or do the best job of formatting it that it can. User's can always undo the format.
- if we choose to do a pre-analysis phase to check for errors ourself, this should be done in the common_server file. The _format impl there can pipe the source through the `analyzer` instance, and then (if it looks good) pipe that through the AS's format call.

lib/src/api_classes.dart
lukechurch: We could do - It's not clear to me it's all that much simpler, given that clients already have source edits implemented for other reasons, and on large files it may be marginally slower. I can go either way here, what would you prefer, server side edit?

lib/src/common_server.dart
lukechurch: ok

lib/src/analysis_server.dart
lukechurch: I agree.

For now, I'd like to pipe it through a pre-analysis check. I'll move the location of the check into common_server.

lib/src/api_classes.dart
lukechurch: Yeah, done.

lukechurch:
Thanks @devoncarew I've adjusted the model to apply the source edits on the server and to move the 'doesn't parse' protection to common_server. I'm not entirely convinced that the edit model is correct under multiple sequential edits, though it is the same as the code mirror does. I'll verify before merging.

Does this address your concerns?

lib/src/api_classes.dart
```diff
@@ -186,6 +186,16 @@ class CandidateFix {
   CandidateFix(this.message, this.edits);
 }
 
+/*
+ * Represents a reformatting of the code.
+ */
+class FormatResponse {
+  String newString;
```
devoncarew: final?

lib/src/api_classes.dart
```diff
@@ -186,6 +186,16 @@ class CandidateFix {
...
+  int offset;
```
devoncarew: final?

lib/src/analysis_server.dart
```diff
@@ -61,16 +61,27 @@ class AnalysisServerWrapper {
...
-      return new FixesResponse(fixes.fixes);
+      return new api.FixesResponse(fixes.fixes);
+    });
+  }
+
+  Future<api.FormatResponse> format(String src, int offset) async {
+    var results = _formatImpl(src, offset);
+    return results.then((editResult) {
+      String editSrc = src;
+      editResult.edits.forEach((edit) {
```
devoncarew: Does the order here matter? Should the edits be sorted by offset, and then applied from back to front?

devoncarew:
Awesome, lgtm; thanks for making the changes.

lib/src/api_classes.dart
lukechurch: done

lib/src/api_classes.dart
lukechurch: done

lib/src/analysis_server.dart
lukechurch: Yes, I have been concerned about this. Because of an underlying bug in AS it's been difficult to verify. But, I think they should be applied back to front. I'll add this. Thanks

lukechurch:
@devoncarew Thanks for the review. 
I'm going to bring this in, to minimise the merge conflicts. I'm currently running extended tests offline. I'll sync the test scripts and the unit tests for format in a follow up PR.

PR #144, opened 2015-04-18
improve compiler error messages

devoncarew:
Improve compiler error messages

- try and filter out dart2js hints, as well as errors/warnings on files that are not the user's source
- when printing locations for errors, only include line numbers for the user's source (line numbers in sdk errors don't help the user)

@lukechurch


lukechurch:
lgtm - could you fix up the build failure before merge?

devoncarew:
fixed -

Issue #77, opened 2015-03-16
improve documentation about the REST API


devoncarew:
How to use it, the maturity level (will it change?), a simple example, ...

Possibly just in the readme?

lukechurch:
Yeah... I'll do this, probably just putting an example on the homepage with how to get start against the new generated clients.

devoncarew:
Closing this one for docs; will add another one for an example.

PR #152, opened 2015-04-19
add documentation for the service calls; regen the library

devoncarew:
Partial fix for #77.

- annotate the rest calls with docs
- re-gen the client library for the protocol


devoncarew:
Also, change to using the `files` generation option on the rpc generator rather than the `package` option. The library now generates as `dartservices.dart` instead of `v1.dart` (which is a nice improvement).

Issue #147, opened 2015-04-19
test coverage for formatter


devoncarew:
possibly mentioned in another issue, but I thought I'd file an issue for it

devoncarew:
done

PR #153, opened 2015-04-19
End to end format test

lukechurch:
Addresses #147

### DIFF
--- a/test/common_server_test.dart
+++ b/test/common_server_test.dart
@@ -213,7 +213,7 @@ void defineTests() {
     });
 
     test('format', () async {
-      var json = {'source': badFormatCode, 'offset': 0};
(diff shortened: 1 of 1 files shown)


test/common_server_test.dart
```diff
@@ -189,6 +211,15 @@ void defineTests() {
       expect(response.status, 200);
       expect(data['count'], 1);
     });
+
+    test('format', () async {
+      var json = {'source': badFormatCode, 'offset': 0};
```
devoncarew: Suggest you change this to _not_ send in an offset, and add another test that does send in an offset, and verifies that the cursor position is updated correctly.

lukechurch:
@devoncarew sg.

PR #156, opened 2015-04-21
fixes test; formatter test

### LINKED ISSUE #148: test coverage for `fix` service call

devoncarew:
@lukechurch 

- add a common server test for `fixes` (fix #148)
- add more coverage for the format call (w/ a position arg)
- fix a bug w the format impl

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -77,7 +77,7 @@ class AnalysisServerWrapper {
 
   Future<api.FormatResponse> format(String src, int offset) async {
     var results = _formatImpl(src, offset);
-    return results.then((editResult) {
+    return results.then((EditFormatResult editResult) {
       String editSrc = src;
       List<SourceEdit> edits = editResult.edits;
       edits.sort((e1, e2) => -1 * e1.offset.compareTo(e2.offset));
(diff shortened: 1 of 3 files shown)


lib/src/analysis_server.dart
```diff
@@ -77,7 +77,7 @@ class AnalysisServerWrapper {
 
   Future<api.FormatResponse> format(String src, int offset) async {
     var results = _formatImpl(src, offset);
-    return results.then((editResult) {
+    return results.then((EditFormatResult editResult) {
```
lukechurch: I prefer leaving this as en explicit type, there are too many levels of indirection to be able know what this thing is going to give my reading the code (how many different forms of result element we have in this code (!)). Otherwise it gets really unclear what this thing is yielding without a lot of hovering over.

lukechurch:
Thanks @devoncarew.

lib/src/analysis_server.dart
devoncarew: I added the `EditFormatResult` type annotation; I think you're saying that it aids readability, and we should keep it? If so, thats my thinking as well :)

lukechurch: Haha. Yes, I think I'm in strong agreement with you, and misreading the diff.

lukechurch:
LGTM, pls merge as soon as build has completed.

Issue #140, opened 2015-04-16
Displayed signature in docs is over-specific


sethladd:
Consider this code:

`Future doThing() => new Future.value(42);`

The docs will show:

`doThing() --> Future<dynamic>`

Ideally, we would not display `<dynamic>` here.

lukechurch:
I've fixed this in the server for now, in discussions with Analzyer team about whether we can fix it deeper into the library.

PR #160, opened 2015-04-22
Future<dynamic> -> Future

lukechurch:
Address #140 

This is a temporary solution, whilst I discuss with the Analyzer team whether we can make the change deeper in the implementation.

If we can't, we'll need to replace print formatters for the Analyzer which is a larger task. For now, lets munge the strings.

TBR @devoncarew

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -140,7 +140,7 @@ class Analyzer {
 
         // Name and description.
         if (element.name != null) info['name'] = element.name;
-        info['description'] = '${element}';
+        info['description'] = '${_pretifyElement(element)}';
         info['kind'] = element.kind.displayName;
 
         // Only defined if there is an enclosing class.
@@ -154,7 +154,8 @@ class Analyzer {
         if (element is ExecutableElement) {
           List<String> list = [];
           ExecutableElement el = element;
-          el.parameters.forEach((par) => list.add('${par}'));
+          el.parameters.forEach((par)
+            => list.add('${_pretifyElement(par)}'));
           info['parameters'] = list;
         }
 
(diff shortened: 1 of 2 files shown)


lib/src/analyzer.dart
```diff
@@ -206,6 +208,19 @@ class Analyzer {
 
     return null;
   }
+
+  // TODO(lukechurch): Determine whether we can change this in the Analyzer.
+  static String _pretifyElement(Element e) {
```
devoncarew: `pretify` ==> `prettify`

lib/src/analyzer.dart
```diff
@@ -206,6 +208,19 @@ class Analyzer {
...
+    String returnString = "${e}";
+    returnString = returnString.replaceAll("Future<dynamic>", "Future");
+    return returnString;
+  }
+
+  static String _pretifyType(DartType dt) {
```
devoncarew: `pretify` ==> `prettify`

lib/src/analyzer.dart
```diff
@@ -206,6 +208,19 @@ class Analyzer {
 
     return null;
   }
+
+  // TODO(lukechurch): Determine whether we can change this in the Analyzer.
+  static String _pretifyElement(Element e) {
+    String returnString = "${e}";
+    returnString = returnString.replaceAll("Future<dynamic>", "Future");
```
devoncarew: Things could be specified as `Future<dynamic>` though. What we don't want is things specified as `Future` to be printed as `Future<dynamic>`.

devoncarew: @keertip, can you point us to what you did in dartdoc to address this?

PR #170, opened 2015-04-28
Update load_driver.dart

lukechurch:
Support lower than 1 qps


lukechurch:
@googlebot Thanks for your help. I signed it!

PR #177, opened 2015-05-03
Upgrade RPC library

### LINKED ISSUE #169: rev appengine rependency
We're behind in the appengine and rpc dependencies, and should rev to the latest. @lukechurch 

```
appengine 0.2.6+3 (0.3.0 available)
rpc 0.2.0 (0.4.1 available)
```

lukechurch:
@wibling PTAL - This is the transition from RPC v2 -> v4

If I don't put in the no-arg ctors, the generator returns failures. This degrades the readability of the API quite a bit, as in a lot of cases these result carrying classes don't make sense without arguments. Is there anything we could do here to avoid the need for these empty ctors?

Fixes #169

### DIFF
--- a/lib/services_gae.dart
+++ b/lib/services_gae.dart
@@ -57,7 +57,7 @@ class GaeServer {
         new GaeCounter());
     // Enabled pretty printing of returned json for debuggability.
     apiServer =
-        new rpc.ApiServer(_API, prettyPrint: true)..addApi(commonServer);
+        new rpc.ApiServer(apiPrefix: _API, prettyPrint: true)..addApi(commonServer);
   }
 
   Future start() => ae.runAppEngine(requestHandler);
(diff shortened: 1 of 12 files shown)


lib/src/api_classes.dart
```diff
@@ -13,7 +13,8 @@ import 'package:rpc/rpc.dart';
 class AnalysisResults {
   final List<AnalysisIssue> issues;
 
-  AnalysisResults(this.issues);
+  AnalysisResults() : this.byIssues([]);
+  AnalysisResults.byIssues(this.issues);
```
devoncarew: Here and for the other ctors, what about `from` instead of `by`? `AnalysisResults.byIssues()` ==> `AnalysisResults.fromIssues()`

devoncarew: (assuming we need to keep the two ctors)

lukechurch:
@devoncarew Sure, happy to rename to that if we do end up needing to keep both ctors.

lib/services_gae.dart
```diff
@@ -57,7 +57,7 @@ class GaeServer {
         new GaeCounter());
     // Enabled pretty printing of returned json for debuggability.
     apiServer =
-        new rpc.ApiServer(_API, prettyPrint: true)..addApi(commonServer);
+        new rpc.ApiServer(apiPrefix: _API, prettyPrint: true)..addApi(commonServer);
```
wibling: Just to let you know. There might be a small overhead in enabling prettyPrint according to Lasse.

wibling:
As the code is right now there is no way to avoid the zero-arg constructors. That said I can see that it worked in the past (I suspect there might have been a bug then :)) and I think I can make it work again for classes that are only used as responses. Basically when the class is used as a response I don't need to be able to create an instance of it. That requirement is only for request classes.
There is some complications in the way the code is structured to fix this, but let me take a look and see what is the simplest solution.

wibling:
Btw. lgtm.

PR #178, opened 2015-05-03
Only compute errors once

lukechurch:
@devoncarew TBR

These errors get computed twice on lines 66, 67 and 71, 72 in the original src.


devoncarew:
Ah, good catch!

PR #179, opened 2015-05-04
Basic multiple files analysis

lukechurch:
@devoncarew PTAL

This adds the ability to analyse multiple dart files via a new EP 'analyzeMulti'. It doesn't link that through to the RPC APIs yet.

It cleans a bit more data betweens calls, but I don't think this will cause a measurable performance impact.


lib/src/analyzer.dart
```diff
@@ -90,13 +105,65 @@ class Analyzer {
...
+      _context.applyChanges(changeSet);
+
+      List<AnalysisErrorInfo> errorInfos = [];
+      sourcesList.forEach((s) {
+        _context.computeErrors(s);
+        errorInfos.add(_context.getErrors(s));
+      });
+
+      List<_Error> errors = errorInfos
+      .expand((AnalysisErrorInfo info) {
```
devoncarew: indent += 2

lib/src/api_classes.dart
```diff
@@ -55,6 +57,24 @@ class SourceRequest {
...
+      description: 'Map of names to sources.')
+  Map<String, String> sources;
+
+  @ApiProperty(
+      description: 'An optional offset into the source code.')
+  int offset;
+
+  @ApiProperty(
+      description: 'Name of file that the offset is in')
+  String offsetName;
```
devoncarew: Having the `offset` and `offsetName` be separate fields seems a little awkward. `offset` could be a `String`, `int` tuple?

devoncarew:
lgtm, some comments

lib/src/api_classes.dart
lukechurch: I'll add a tiny 'Location' class to handle this.

lib/src/analyzer.dart
lukechurch: done

lukechurch:
Thanks @devoncarew with simple tests I couldn't get a difference that was measurable. I'll try again with a larger test set, but it looks like it's going to be fine wrt performance.

I've addressed the review comments.

lukechurch:
I will add an additional benchmark for multi-files in a separate PR.

lukechurch:
@googlebot I am the author of these commits. The build break they caused is also entirely my fault.

PR #186, opened 2015-05-07
add a /version service call

### LINKED ISSUE #185: service call to get dart / sdk version
Related to https://github.com/dart-lang/dart-pad/issues/403, we need a service call that will tell us what version of dart2js and the analyzer we're running (a single version for both).

devoncarew:
Fix #185 

@lukechurch 

Should the service call be called `/sdkVersion`? `/version`? I don't want people to confuse the sdk version and a dartpad version. Perhaps call it `/version` and have it return a few different versions in the result object.

### DIFF
--- a/doc/generated/dartservices.dart
+++ b/doc/generated/dartservices.dart
@@ -566,7 +566,7 @@ class DartservicesApi {
    * If the used [http.Client] completes with an error when making a REST call,
    * this method  will complete with the same error.
    */
-  async.Future<VersionResponse> sdkVersion() {
+  async.Future<VersionResponse> version() {
     var _url = null;
     var _queryParams = new core.Map();
     var _uploadMedia = null;
@@ -576,7 +576,7 @@ class DartservicesApi {
 
 
 
-    _url = 'sdkVersion';
+    _url = 'version';
 
     var _response = _requester.request(_url,
                                        "GET",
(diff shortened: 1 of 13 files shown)


lukechurch:
@devoncarew I like the idea of having it be /version and return a result with versions of all the subcomponents - the SDK, the compiler, the analyzer, the Analysis Server, the formatter and Dart-Services itself.

doc/generated/dartservices.dart
```diff
@@ -553,6 +553,41 @@ class DartservicesApi {
...
+   *
+   * Completes with a [VersionResponse].
+   *
+   * Completes with a [commons.ApiRequestError] if the API endpoint returned an
+   * error.
+   *
+   * If the used [http.Client] completes with an error when making a REST call,
+   * this method  will complete with the same error.
+   */
+  async.Future<VersionResponse> sdkVersion() {
```
lukechurch: AS discussed lets change this to version

doc/generated/dartservices.dart
```diff
@@ -553,6 +553,41 @@ class DartservicesApi {
...
+    var _url = null;
+    var _queryParams = new core.Map();
+    var _uploadMedia = null;
+    var _uploadOptions = null;
+    var _downloadOptions = commons.DownloadOptions.Metadata;
```
lukechurch: What does this do?

lukechurch:
I think I'd suggest the following mapping:

- dart2JS version
- Libraries SDK version
- appengine SDK version
- Analyzer version
- Analysis Server version

Once things stabilise hopefully we'll get to a point where all of these are in sync, but the numbers will still be different.

devoncarew:
I'll update to rename to `/version`, and add some fields to the response.

I have concerns about identifying the versions of all the individual components. In an ideal world, we'll be `1.10.0` across the board for dart2js, the analyzer, and the analysis server. The client should only care that the backend is compatible with a particular sdk version of dart. Mostly I want to make sure that we maintain a distinction between the sdk version that dartpad is on and any version number of the actual backend or frontend - revs to the tool vs. the sdk it's compatible with.

lukechurch:
ok - sg, we can drop the version reporting the sub components. I would like to have a 'dart-services version' that we can use, it's very helpful for sanity checking that deployments are serving correctly, but we can do that just as a string constant for now.

devoncarew:
Do you know, is the appengine context version different from the version of the sdk that the app is running on?

devoncarew:
`ModulesService.currentVersion`

devoncarew:
OK, updated:

- renamed the call to `/version`
- the version reponse now has four fields: `sdkVersion`, `runtimeVersion`,
`appEngineVersion`, and `servicesVersion`
- there's a `version.dart` file which has a services version in it. This will be useful to know what's deployed and to have something for user's to file bugs against. We'll have to figure out an easy process to maintain it w/ releases. Feedback welcome!

devoncarew:
`sdkVersion` is the main one, in terms of that's the SDK that dartpad currently supports.

doc/generated/dartservices.json
```diff
@@ -208,6 +208,28 @@
...
     }
    }
+  },
+  "VersionResponse": {
+   "id": "VersionResponse",
+   "type": "object",
+   "properties": {
+    "sdkVersion": {
+     "type": "string",
+     "description": "The version of the Dart SDK that DartPad is compatible with. This will be a semver string."
```
lukechurch: s/DartPad/DartServices

doc/generated/dartservices.json
```diff
@@ -437,6 +459,17 @@
    "response": {
     "$ref": "DocumentResponse"
    }
+  },
+  "version": {
+   "id": "CommonServer.version",
+   "path": "version",
+   "httpMethod": "GET",
+   "description": "Return the current SDK version for DartPad.",
```
lukechurch: s/DartPad/DartServices

lib/src/api_classes.dart
```diff
@@ -237,3 +237,30 @@ class SourceEdit {
     return "$pre$replacement$post";
   }
 }
+
+/// The response from the `/version` service call.
+class VersionResponse {
+  @ApiProperty(
+      description: 'The Dart SDK version that DartPad is compatible with. This '
```
lukechurch: s/DartPad/DartServices

lib/src/common_server.dart
```diff
@@ -190,6 +198,12 @@ class CommonServer {
     return _document(source, offset);
   }
 
+  @ApiMethod(
+      method: 'GET',
+      path: 'version',
+      description: 'Return the current SDK version for DartPad.')
```
lukechurch: s/DartPad/DartServices

lukechurch:
lgtm with naming nits

doc/generated/dartservices.json
devoncarew: done

doc/generated/dartservices.json
devoncarew: done

lib/src/api_classes.dart
devoncarew: done

lib/src/common_server.dart
devoncarew: done

PR #189, opened 2015-05-07
compile in checked mode; return source maps

### LINKED ISSUE #129: Support checked mode compilation
To allow for demonstrations of the difference. 
Per: https://github.com/dart-lang/dart-pad/issues/310

### LINKED ISSUE #130: return generated source maps
When compiling, some clients may want the generated source maps. To reduce cost for clients tha don't want them, this should be an opt-in feature (optional param). We'd need to change the return type of the `compile` call to return both the compiled JS results and the generated sourcemap file, if any.

devoncarew:
fix #130, fix #129 

- return source maps for `/compile` (when requested, to save round trip time)
- compile in checked mode (when requested)

@lukechurch

### DIFF
--- a/doc/generated/dartservices.dart
+++ b/doc/generated/dartservices.dart
@@ -117,7 +117,7 @@ class DartservicesApi {
    * If the used [http.Client] completes with an error when making a REST call,
    * this method  will complete with the same error.
    */
-  async.Future<CompileResponse> compile(SourceRequest request) {
+  async.Future<CompileResponse> compile(CompileRequest request) {
     var _url = null;
     var _queryParams = new core.Map();
     var _uploadMedia = null;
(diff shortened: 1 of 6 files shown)


lukechurch:
@devoncarew it's a bit more complicated than this due to the use of caching. I'll take this PR from here and make it good with our memcache scheme.

Am I correct in thinking that source maps are going to be different for checked and unchecked mode?

Do you know if doing this is going to break existing clients due to the change to using CompileRequests rather than SourceRequests?

devoncarew:
@lukechurch, sounds good.

Yes, source maps will be different for checked vs unchecked. You'll probably want to append the optional params to the memcache key, and store the full json response as the memcache value. I totally forgot about the caching :(

> Do you know if doing this is going to break existing clients due to the change to using CompileRequests rather than SourceRequests?

I don't _think_ so, but it would be worth checking. The json should be identical.

Issue #190, opened 2015-05-09
guard against memcache API errors


devoncarew:
See https://github.com/dart-lang/dart-pad/issues/408.

devoncarew:
fixed

PR #191, opened 2015-05-11
Memcache error protection

lukechurch:
@sgjesse @mkustermann @wibling PTAL

We're still seeing occasional bursts of errors on memcache. I think this is about as defensive as we can - does it look reasonable?

@devoncarew FYI, this addresses #190

### DIFF
--- a/lib/services_gae.dart
+++ b/lib/services_gae.dart
@@ -115,13 +115,27 @@ class GaeServer {
 class GaeCache implements ServerCache {
   Memcache get _memcache => ae.context.services.memcache;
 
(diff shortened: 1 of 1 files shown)


sgjesse:
You can also see the change Martin did for the pub server: https://github.com/dart-lang/pub-dartlang-dart/commit/d287e3322b19a5eb1148b6149edca9dab5810b92.

Maybe adding a wrapper for ignoring errors is better than repeating the same code three times.

lukechurch:
@sgjesse ah yes, I like that solution better, if nothing else it's more readable. Thank you.

PR #197, opened 2015-05-13
upgrade rpc to 0.5.0

devoncarew:
- upgrade `rpc` to 0.5.0
- revert to using ctors with args on response objects
- fix a bug in the `/version` service call, where the appengine version wasn't reported properly

@lukechurch


lukechurch:
Thanks @devoncarew, 

lgtm!

PR #202, opened 2015-05-15
add additional analysis info to the /analyze service call

### LINKED ISSUE #162: add an additional sidecar result to the analysis service call
Related to https://github.com/dart-lang/dart-pad/issues/3. One way of solving this is to have the `/analyze` service call return a full list of the transitive imports. The front-end can then check for `dart:io` and take whatever steps it wants. The current result of `/analyze` is:

```
class AnalysisResults {
  final List<AnalysisIssue> issues;

  AnalysisResults(this.issues);
}
```

We could add a list of transitive imports (and perhaps the direct imports?).

```
final List<String> transitiveImports;
final List<String> directImports;
```

Or:

```
final List<Import> imports;

class Import {
  final String name;
  final bool isDirect;
}
```

devoncarew:
- fix #162, add import info to the `/analyze` service call (to make the client aware of things like a transitive `dart:io` import)
- re-gen the api library (capture the `/analyzeMulti` call and the sidecar analysis info change)
- rename `fullName` to `sourcePath` for clarity

@lukechurch

### DIFF
--- a/doc/generated/dartservices.dart
+++ b/doc/generated/dartservices.dart
@@ -644,12 +644,12 @@ class AnalysisIssue {
 
   core.int line;
 
-  /** deprecated - see sourcePath */
+  /** deprecated - see `sourceName` */
   core.String location;
 
   core.String message;
 
-  core.String sourcePath;
+  core.String sourceName;
 
 
   AnalysisIssue();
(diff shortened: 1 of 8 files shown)


lukechurch:
lgtm, thanks

PR #203, opened 2015-05-15
Support multiple files for completion, fixes

lukechurch:
@devoncarew PTAL This adds support for multiple files in the EPs that use the Analysis Server.

It needs two follow on PRs:

1. Add unit tests for multiple files. The code paths are tested with EPs for single files but needs extending.
2. Adjust the sendOverlay implementation so that it sends the overlays in one Analysis Server message rather than 1 per file.

I'll file issues for these, neither affects the correctness of the functioning of this implementation

### DIFF
--- a/doc/generated/dartservices.dart
+++ b/doc/generated/dartservices.dart
@@ -305,6 +305,46 @@ class DartservicesApi {
   }
 
   /**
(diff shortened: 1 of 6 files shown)


lukechurch:
TBR

devoncarew:
lgtm!

PR #205, opened 2015-05-16
Remove additional CORS settings

lukechurch:
@devoncarew TBR These values are now being set in one of the libraries, which was resulting in them being set twice, and hence causing requests to fail.

With this change the dev deploy of the UI talks successfully with a dev deploy of the services library. Doing more manual testing now.


devoncarew:
Interesting - is this now done by `rpc`?

We might also hit the server with a client running on Safari, which has historically been twitchy w/ CORS settings.

lgtm

PR #206, opened 2015-05-17
Update README.md

devoncarew:
remove the beta label

@lukechurch


lukechurch:
Hmm, I'll look into what is up with the failing test and then merge this. It looks like Travis was having trouble with git pulls before, which is why I didn't see this.

But, all that said. LGTM!

lukechurch:
I think it might be worth saying that it's still early days, and so we do still reserve the right to break the APIs...

lukechurch:
@googlebot don't worry, it'll all be ok

lukechurch:
These tests pass locally, waiting for another build

PR #211, opened 2015-05-20
Fix a race condition in the interaction with the Analysis Server

lukechurch:
@devoncarew - TBR, sorry - This is a hot fix to remove a problematic race condition that is starting to hit our usage by adding a mutex to _impl mechanisms that communicate with the server. After this fix the issue can't be reproduced.

I need to adapt our test procedures so we can catch these in local testing. The better long term fix for this is to schedule these requests onto a queue.

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -37,6 +37,7 @@ var mainPath;
 
 class AnalysisServerWrapper {
   final String sdkPath;
+  bool serverBusy = false;
 
   /// Instance to handle communication with the server.
   _Server serverConnection;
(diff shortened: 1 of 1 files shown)


lib/src/analysis_server.dart
```diff
@@ -284,6 +307,7 @@ class _Server {
     _logStdio('PROCESS FORCIBLY TERMINATED');
     _process.kill();
     Future<int> exitCode = _process.exitCode;
+    isSetup = false;
```
lukechurch: This is needed to make the self-repair mechanism work properly

devoncarew:
Just for clarity, the issue was that we're getting a code completion request while still processing an existing one?

lukechurch:
Yup, the request handler from GAE can fire when we're awaiting. This means that we can get interleaved requests on anything that uses Analysis Server, which we need to serialise.

PR #212, opened 2015-05-25
Add single step deployment tool

lukechurch:
@devoncarew Is there an advantage to having this as a manual grind action or leave it as a series of shell operations? It needs to be a manual step.

### DIFF
--- a/tool/warmup.dart
+++ b/tool/warmup.dart
@@ -15,11 +15,12 @@ const count = 10;
 main(List<String> args) async {
 
   String appPrefix;
-  if (args.length > 0)
+  if (args.length > 0) {
     appPrefix = "${args[0]}.";
-  else
+  } else {
     appPrefix = "";
-
+  }
+  
   // Use an insecure connection for test driving to avoid cert problems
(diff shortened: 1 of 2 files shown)


tool/warmup.dart
```diff
@@ -6,10 +6,26 @@ import 'dart:convert' as convert;
...
+String uri;
+
+const BASE_URI = "dart-services.appspot.com/api/dartservices/v1/";
+//const BASE_URI = "http://localhost:8080/api/dartservices/v1/";
+const count = 10;
 
 main(List<String> args) async {
+
+  String appPrefix;
+  if (args.length > 0)
```
devoncarew: need braces here -

devoncarew:
@lukechurch, I don't see an advantage either way -

To fix the build issue, change `tool/grind.dart` line 24 to `test() => Dart.run('test/all.dart');`.

devoncarew:
The build issue was addressed in https://github.com/dart-lang/dart-services/pull/213 and the fix is now in `master` :)

PR #214, opened 2015-05-29
Harden analysis server integration

lukechurch:
@devoncarew 

This PR makes two changes

- Replace spin locks with a scheduler that can do time outs
- Uses this to deal with the case where the Analysis Server never completes some requests

It also contains a bulk reformat to use the formatter - I should have done in a separate PR, sorry about that, but the impact isn't too bad.

In the very long running test suite these correctly result in the Analysis Server self-healing. The thing that causes it to die appears to be a state dependent issue rather than a single code change issue. I can currently only reproduce the case after around 30 hours of usage - this doesn't fix that problem it just bounds the outage.

I will follow up with the AS team about getting the underlying bugs fixed.

cc: @danrubel

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -34,9 +34,9 @@ final _WARMUP_SRC = "main() { int b = 2;  b++;   b. }";
 final _SERVER_PATH = "bin/_analysis_server_entry.dart";
 
 // Use very long timeouts to ensure that the server has enough time to restart.
-const int _ANALYSIS_SERVER_TIMEOUT_S = 15;
-const int _COMPILE_TIMEOUT_S = 25;
-const int _ANALYZE_TIMEOUT_S = 15;
+final _ANALYSIS_SERVER_TIMEOUT = new Duration(seconds: 15);
+final _COMPILE_TIMEOUT = new Duration(seconds: 25);
+final _ANALYZE_TIMEOUT = new Duration(seconds: 15);
 
 var sourceDirectory;
 var mainPath;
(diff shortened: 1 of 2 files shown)


lib/src/analysis_server.dart
```diff
@@ -32,12 +33,17 @@ final _WARMUP_SRC_HTML = "import 'dart:html'; main() { int b = 2;  b++;   b. }";
 final _WARMUP_SRC = "main() { int b = 2;  b++;   b. }";
 final _SERVER_PATH = "bin/_analysis_server_entry.dart";
 
+// Use very long timeouts to ensure that the server has enough time to restart.
+const int _ANALYSIS_SERVER_TIMEOUT_S = 15;
```
devoncarew: recommend to use `Duration` here instead -

lib/src/analysis_server.dart
```diff
@@ -47,20 +53,18 @@ class AnalysisServerWrapper {
...
     serverConnection = new _Server(this.sdkPath);
+    serverScheduler = new scheduler.TaskScheduler();
   }
 
   Future<api.CompleteResponse> complete(String src, int offset) async =>
-    completeMulti({"main.dart" : src},
-      new api.Location()
-        ..sourceName = "main.dart"
-        ..offset = offset);
+      completeMulti({"main.dart": src}, new api.Location()
```
devoncarew: here and below - perhaps prefer to not use the function shorthand when the function spans across several lines (even if it is only one expression)?

devoncarew:
lgtm! Is the build failure (re: the benchmarks) something you can repo locally, or is it transient?

lib/src/analysis_server.dart
lukechurch: Yeah, good point. Will do.

lib/src/analysis_server.dart
lukechurch: ok.

lukechurch:
I think the build failure is transient, it doesn't occur locally for me. If it repeats on the push to address your other comments I'll dig into it a bit further.

lib/src/analysis_server.dart
lukechurch: done. This feels better and is more consistent with the other methods.

Issue #219, opened 2015-06-04
Leak in analysis


lukechurch:
Analysis performance degrades over time. This causes a progressive slow down over a number of days in response to /analysis calls.

Perf response under a local stress loading looks like this (x axis sequential requests, y axis response time/ms for a local test)

![image](https://cloud.githubusercontent.com/assets/2978182/7991857/23bca420-0ab0-11e5-8b15-8c85b28ed161.png)

Current working hypothesis is a leak in the context model. Looking into it.

lukechurch:
After fix

![image](https://cloud.githubusercontent.com/assets/2978182/7994069/7561d55c-0abf-11e5-96c1-9f4726359a7c.png)

PR #220, opened 2015-06-04
We should be removing sources not deleting them

lukechurch:
@devoncarew PTAL

We were causing sources to accumulate which was causing analysis inefficiency.

Please see #219


devoncarew:
lgtm!

PR #222, opened 2015-06-08
re-gen the services library to capture better gen output

devoncarew:
Re-gen the services library to capture better gen output from the `discoveryapis_generator` library. No functional changes -

@lukechurch


lukechurch:
lgtm, thanks

PR #223, opened 2015-06-08
Defork Analyzer and Dart_style, update server hardware settings

lukechurch:
- Use a new CPU profile
- Defork analyzer and dart_style

Bringing this in for stress testing

### DIFF
--- a/app.yaml
+++ b/app.yaml
@@ -1,14 +1,14 @@
-version: v191-j
+version: v1-10-1
 runtime: custom
 vm: true
 api_version: 1
 
 resources:
-  cpu: 2
+  cpu: 1
   memory_gb: 3
 
 automatic_scaling:
-  min_num_instances: 3
+  min_num_instances: 6
   max_num_instances: 20
   cool_down_period_sec: 60
   cpu_utilization:
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -9,15 +9,9 @@ dependencies:
   analysis_server:
     git:
       url: git://github.com/lukechurch/dart-analysis-server-temp-working.git
-      ref: v1_10_1
-  analyzer:
-    git:
-      url: git://github.com/lukechurch/dart-analyzer-temp-working.git
-      ref: v1_10_1
-  dart_style:
-    git:
-      url: git://github.com/lukechurch/dart_style.git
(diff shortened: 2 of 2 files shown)


app.yaml
```diff
@@ -1,14 +1,14 @@
-version: v191-j
+version: v1-10-1
 runtime: custom
 vm: true
 api_version: 1
 
 resources:
-  cpu: 2
+  cpu: 1
```
devoncarew: If I recall correctly, we had overhead issues when we went down to one actual core. Are things better now?

pubspec.yaml
```diff
@@ -9,15 +9,9 @@ dependencies:
...
-    git:
-      url: git://github.com/lukechurch/dart-analyzer-temp-working.git
-      ref: v1_10_1
-  dart_style:
-    git:
-      url: git://github.com/lukechurch/dart_style.git
-      ref: v1_10_1
+      ref: v1_11_0
+  analyzer: '>=0.25.1-alpha.1 <0.26.0'
+  dart_style: '>=0.1.7 <0.2.0'
```
devoncarew: +1000!

Do we need an explicit reference to `dart_style`? I don't think we use the package directly, just indirectly through the AS.

devoncarew:
I see this locally now:

```
Resolving dependencies... 
Git error. Command: git rev-list --max-count=1 v1_11_0
fatal: ambiguous argument 'v1_11_0': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

app.yaml
lukechurch: Yes. With the addition of the time out the thing that we were using multiple CPU cores to avoid is resolved. I did some testing, and with that we get considerably better execution density having more single core instances than fewer multi-core instances. This keeps the scheduler queues as short as possible.

We will want to revisit this if we do move to an isolate model.

pubspec.yaml
lukechurch: nope, we should be able to drop this. At this stage just decreasing the number of changes. Assuming that all the of the detailed tests pass, I'll get rid of this explicit dependency shortly.

PR #225, opened 2015-06-09
Add deployment instructions

lukechurch:
Addresses #19


lukechurch:
@googlebot `I signed it!`

PR #228, opened 2015-06-14
Add missing perform analysis task

lukechurch:
TBR @devoncarew 

This is needed to fix a slow memory leak that is causing a progressive decay in server performance over a number of days.

I've done a 250K test, and it improves performance by decreasing the stress on the GC. With this change memory footprint for the Analyzer is stable at 300Mb. Without it, it grows to system exhaustion.

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -24,6 +24,8 @@ import 'pub.dart';
 
 Logger _logger = new Logger('analyzer');
 
+final Duration _MAX_ANALYSIS_DURATION = new Duration(seconds: 5);
+
 class MemoryResolver extends UriResolver {
   Map<String, Source> sources = {};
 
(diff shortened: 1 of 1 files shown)


lib/src/analyzer.dart
```diff
@@ -123,6 +123,7 @@ class Analyzer {
       sourcesList.forEach((s) => changeSet.removedSource(s));
       _resolver.clear();
       _context.applyChanges(changeSet);
+      _context.performAnalysisTask();
```
devoncarew: You might want to move this above line 91, before we calculate the issues. In other projects, I've used this in conjunction with `hasMoreWork` (https://github.com/google/tuneup.dart/blob/master/lib/src/check_command.dart#L60).

lukechurch:
@devoncarew Thanks for the pointer - I think in our case we need to do it in both places, I've added a helper method with some basic timeout protection.

Issue #229, opened 2015-06-24
upgrade to sdk upgrade to 1.11.0


devoncarew:
It'll be released shortly. We'll need to rev. the analyzer, compiler_unsupported, and the analysis server version.

devoncarew:
How's this looking? We're on `Dart SDK 1.10.1` currently, and SDK 1.11.1 is the latest stable sdk release.

lukechurch:
Testing looks good. I'm routing all the traffic to 1.11 now. I shouldn't have close this issue before then.

Have you upgraded the compiler package to 1.11.1, or want to wait to 1.12?

lukechurch:
![image](https://cloud.githubusercontent.com/assets/2978182/8565363/8af9a312-2555-11e5-9448-53f93f9e32e2.png)

devoncarew:
I'll investigate 1.11.1 - the repo change may have made this more difficult.

devoncarew:
1.11.1 of compiler_unsupported is available now

lukechurch:
SG. I'll rev in a few hours, close out this issue for now that we have a tracking issue for migrating to 1.11.1.

PR #230, opened 2015-06-25
upgrade to 1.11

devoncarew:
fix https://github.com/dart-lang/dart-services/issues/229

@lukechurch


lukechurch:
If only it were this simple ;)

The usual pre-release stress testing is throwing up issues that I'm hunting down. I'm going to need a couple of days more before we can upgrade the analyzer.

PR #240, opened 2015-07-10
George export api

Georgehe4:
Merged into server as separate API.

@lukechurch @devoncarew


doc/generated/_dbservices.json
```diff
@@ -0,0 +1,77 @@
+{
+ "kind": "discovery#restDescription",
+ "etag": "f86deff6acf8d14c2a21bf74ed893efa106783e6",
+ "discoveryVersion": "v1",
+ "id": "_dbservices:v1",
+ "name": "_dbservices",
```
lukechurch: Lets go for something a bit more descriptive here. These aren't really database services, they're very specific. Something like _dartpadSupportServices?

lib/services_gae.dart
```diff
@@ -44,43 +45,43 @@ class GaeServer {
...
     _logger.level = Level.ALL;
 
     discoveryEnabled = false;
-    commonServer = new CommonServer(
-        sdkPath,
-        new GaeServerContainer(),
-        new GaeCache(),
-        new GaeSourceRequestRecorder(),
-        new GaeCounter());
+    databaseServer = new DbServer();
```
lukechurch: As above. it's not a Database server, we need a more specific name, something like FileRelayServer

lib/services_gae.dart
```diff
@@ -44,43 +45,43 @@ class GaeServer {
...
+    request.response.headers.add(
+        'Access-Control-Allow-Methods', 'DELETE, POST, OPTIONS');
     request.response.headers.add('Access-Control-Allow-Headers',
         'Origin, X-Requested-With, Content-Type, Accept');
 
     // Explicitly handle an OPTIONS requests.
     if (request.method == 'OPTIONS') {
       var requestedMethod =
-        request.headers.value('access-control-request-method');
+          request.headers.value('access-control-request-method');
```
lukechurch: I'm assuming that these ws changes are form dart-style?

lib/services_gae.dart
```diff
@@ -93,20 +94,23 @@ class GaeServer {
       // the _parseRequest method to determine content-type and dispatch to e.g.
       // a plain text handler if we want to support that.
       var apiRequest = new rpc.HttpApiRequest.fromHttpRequest(request);
-      apiServer.handleHttpApiRequest(apiRequest)
-        .then((rpc.HttpApiResponse apiResponse) {
-          return rpc.sendApiResponse(apiResponse, request.response);
+      apiServer
+          .handleHttpApiRequest(apiRequest)
```
lukechurch: I'd like the ws changes in a seperate PR please. Otherwise this is going to be too difficult to review.

I don't mind if you do the reformat before or after the addition of the new API. Your call.

lib/services_server.dart
```diff
@@ -18,6 +18,7 @@ import 'package:shelf_cors/shelf_cors.dart' as shelf_cors;
 import 'package:shelf_route/shelf_route.dart';
 
 import 'src/common_server.dart';
+import 'src/db_server.dart';
```
lukechurch: db name.

lib/services_server.dart
```diff
@@ -29,6 +30,7 @@ void main(List<String> args) {
   parser.addOption('port', abbr: 'p', defaultsTo: '8080');
   parser.addOption('dart-sdk');
   parser.addFlag('discovery');
+  parser.addFlag('database');
```
lukechurch: db name.

lib/services_server.dart
```diff
@@ -39,18 +41,26 @@ void main(List<String> args) {
...
   if (result['discovery']) {
     var serverUrl = result['server-url'];
-    EndpointsServer.generateDiscovery(sdkDir.path, serverUrl).then((doc) {
-      print(doc);
-      exit(0);
-    });
+    if (result['database']) {
+      EndpointsServer
+          .generateDatabaseDiscovery(sdkDir.path, serverUrl)
+          .then((doc) {
```
lukechurch: Please factor this repeated code out into a tiny helper method. You can declare it inline to this method.

lib/services_server.dart
lukechurch: I'm going to stop commenting on this from here - please change in all of the places in the code

lib/services_gae.dart
```diff
@@ -44,43 +45,43 @@ class GaeServer {
...
+      ..addApi(databaseServer);
   }
 
   Future start() => ae.runAppEngine(requestHandler);
 
   void requestHandler(io.HttpRequest request) {
-    request.response.headers.add('Access-Control-Allow-Methods',
-        'POST, OPTIONS');
+    request.response.headers.add(
+        'Access-Control-Allow-Methods', 'DELETE, POST, OPTIONS');
```
lukechurch: You're adding DELETE here but not below in the router. Are we using this verb?

lib/services_server.dart
```diff
@@ -108,22 +128,18 @@ class EndpointsServer {
...
-      .addMiddleware(_createCustomCorsHeadersMiddleware());
+        .addMiddleware(logRequests())
+        .addMiddleware(_createCustomCorsHeadersMiddleware());
 
     routes = router()
-        ..get('/', printUsage)
-        ..add('/api', ['GET', 'POST', 'OPTIONS'], _apiHandler,
-              exactMatch: false);
+      ..get('/', printUsage)
+      ..add('/api', ['GET', 'POST', 'OPTIONS'], _apiHandler, exactMatch: false);
```
lukechurch: Do we need DELETE here?

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+import 'package:crypto/crypto.dart' as crypto;
+import 'dart:convert' as convert;
+
+// This class defines the interface that the server provides.
+@ApiClass(name: '_dbservices', version: 'v1')
+class DbServer {
+  DbServer();
+
+  @ApiMethod(method: 'POST', path: 'export')
+  Future<KeyContainer> returnKey(DataSaveObject data) {
```
lukechurch: The names of functions should say what they do - in this case maybe it's 'storeTempData'?

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+    GaeSourceRecord record = new GaeSourceRecord.FromDSO(data);
+    sha1(record);
+    db.dbService.commit(inserts: [record]).catchError((error, stackTrace) {
+      print('Error recording');
```
lukechurch: This needs to be populated before we can land TBH. PTAL a look at the way that the logging package is used else where in dart-services for examples. If you can't find one, let me know.

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+
+// This class defines the interface that the server provides.
+@ApiClass(name: '_dbservices', version: 'v1')
+class DbServer {
+  DbServer();
+
+  @ApiMethod(method: 'POST', path: 'export')
+  Future<KeyContainer> returnKey(DataSaveObject data) {
+    GaeSourceRecord record = new GaeSourceRecord.FromDSO(data);
+    sha1(record);
```
lukechurch: sha1 should be a pure function - it's a computation of a number, so it shouldn't have side effects. So we'll need to adjust this a little.

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+    db.dbService.commit(inserts: [record]).catchError((error, stackTrace) {
+      print('Error recording');
+    });
+    return new Future.value(new KeyContainer.FromKey(record.UUID));
+  }
+
+  @ApiMethod(method: 'DELETE', path: 'return')
```
lukechurch: Ah nice use of the verb.

lukechurch: Lets not use return as a verb, that's going to make the clientside code use something that is a keyword in most languages in a strange way.

How about "pullExportedData"?

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+  Future<DataSaveObject> returnContent({String key}) {
```
lukechurch: name as per above

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+    //TODO: Query for, and delete the specified object.
+    var database = ae.context.services.db;
+    GaeSourceRecord record;
+    var query = database.query(GaeSourceRecord)..filter('UUID =', key);
+    return query.run().toList().then((List result) {
```
lukechurch: Can you convert this to async/await style?

E.g. var result = await query.run().toList()

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+      if (result.isEmpty) return new DataSaveObject();
```
lukechurch: Add logging pls.

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+    sha1(record);
+    db.dbService.commit(inserts: [record]).catchError((error, stackTrace) {
+      print('Error recording');
+    });
+    return new Future.value(new KeyContainer.FromKey(record.UUID));
+  }
+
+  @ApiMethod(method: 'DELETE', path: 'return')
+  Future<DataSaveObject> returnContent({String key}) {
+    //TODO: Query for, and delete the specified object.
```
lukechurch: You're doing this with the database.commit(deletes:....... call below right?

lib/src/db_server.dart
lukechurch: Should be new Future.value(.... ?

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+          new DataSaveObject.FromData(record.dart, record.html, record.css));
+    });
+  }
+}
+
+/*
+ * This is the schema for source code storage
+ */
+@db.Kind()
+class GaeSourceRecord extends db.Model {
```
lukechurch: Does this need to be here? It's the same as the other class?

lukechurch: Please can we use a different model name so it doesn't conflict with the other API

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+
+class KeyContainer {
+  String key;
+  KeyContainer();
+  KeyContainer.FromKey(String key) {
+    this.key = key;
+  }
+}
+
+// SHA1 set the id
```
lukechurch: Sha functions should be pure and return the SHA, rather than applying the side effect within the function

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+void sha1(GaeSourceRecord record) {
+  crypto.SHA1 sha1 = new crypto.SHA1();
+  convert.Utf8Encoder utf8 = new convert.Utf8Encoder();
+  sha1.add(utf8.convert('blob \n ' + record.html + record.css + record.dart));
```
lukechurch: Please don't add strings together in Dart.

You can make repeated calls to .add, or you could use a string continuation style "blob 'n ${record.html} ${record.css} ... etc

lib/src/db_server.dart
```diff
@@ -0,0 +1,109 @@
...
+
+  DataSaveObject.FromData(String dart, String html, String css, {String UUID}) {
+    this.dart = dart;
+    this.html = html;
+    this.css = css;
+    this.UUID = UUID;
+  }
+}
+
+class KeyContainer {
```
lukechurch: Do we really need this? Could we just mark the return type of the function to be a string?

lukechurch:
Thanks @Georgehe4 this is a great start, and we're actually pretty close I think, don't worry about the number of comments.

Please can you separate out the formatting changes that'll make the review go much smoother. I would suggest doing them in a separate PR, the review should be pretty painless, we can then merge that into this one.

Before anything goes into master, and therefore into the serving product, it does need to have a test case written and logging the behaviour. 

If you could make the changes here and then send it back to me for another look, that'd be great. Thanks!

lib/src/db_server.dart
Georgehe4: Can you explain what you mean here?

lib/src/db_server.dart
Georgehe4: String does not work unfortunately :(

lib/src/db_server.dart
Georgehe4: Yeah... there were differences in access functions found in each library (db and gae). 
This seems odd using two libraries to interact with the same data source but it worked in my local test.

PR #241, opened 2015-07-12
Export - Final

Georgehe4:
Reset the commit history to the commit prior to formatting. Will ping when suggested changes in last PR #240 are completed.


Georgehe4:
@lukechurch

lukechurch:
@Georgehe4 Thanks for this. I think there are still a bunch more changes from the previous PR to address? Are you still working on it, or do you need me to move the comments over?

In terms of error recording, what I mean is that these errors and stack traces need to go in the log. Take a look at https://github.com/dart-lang/dart-services/blob/master/lib/services_gae.dart#L102

Georgehe4:
@lukechurch 
Correct, there are still some things to change such as logging. I will ping you if I finish up these changes or if questions come up. Some of the changes bring up questions that I would like to talk with you on Monday about if you have time.

lukechurch:
Sounds good. I'm drop you a mail with my schedule for Monday

Georgehe4:
Ready for review.
@lukechurch

lib/src/dartpad_support_server.dart
```diff
@@ -0,0 +1,122 @@
...
+      .catchError((e) {
+      _logger.severe("Error while recording export ${e}");
+      throw(e);
+    });
+    _logger.info("PERF: Recorded Export with ID ${record.UUID}");
+    return new Future.value(new KeyContainer.FromKey(record.UUID));
+  }
+  
+  @ApiMethod(method:'DELETE', path:'pullExportData')
+  Future<DataSaveObject> pullExportContent({String key}) async {
```
lukechurch: Lets call this 'PadSaveObject'. Data is too generic.

lib/src/dartpad_support_server.dart
```diff
@@ -0,0 +1,122 @@
...
+  
+  GaeExportRecord.FromDSO(DataSaveObject dso) {
+    this.dart = dso.dart;
+    this.html = dso.html;
+    this.css = dso.css;
+    this.UUID = dso.UUID;
+  }
+}
+
+class DataSaveObject {
```
lukechurch: Lets put a creation time on this object. It'll make it easier to GC the things that don't deleted because of datastore write failures.

lib/src/dartpad_support_server.dart
```diff
@@ -0,0 +1,122 @@
...
+  @ApiMethod(method:'POST', path:'export') 
+  Future<KeyContainer> export(DataSaveObject data){
+    GaeExportRecord record = new GaeExportRecord.FromDSO(data);
+    sha1SetUUID(record);
+    db.dbService.commit(inserts: [record])
+      .catchError((e) {
+      _logger.severe("Error while recording export ${e}");
+      throw(e);
+    });
+    _logger.info("PERF: Recorded Export with ID ${record.UUID}");
```
lukechurch: I use 'PERF' to indicate Performance log entries. These aren't really performance markers, lets drop the 'PERF" from the front.

lib/src/dartpad_support_server.dart
```diff
@@ -0,0 +1,122 @@
...
+    _logger.info("PERF: Deleted Export with ID ${record.UUID}");
+    return new Future.value(new DataSaveObject.FromData(record.dart, record.html, record.css));
+  }
+}
+
+/*
+ * This is the schema for source code storage
+ */
+@db.Kind()
+class GaeExportRecord extends db.Model {
```
lukechurch: This needs a high precision time stamp associated with it. See 

  @db.IntProperty()
  int epochTime; 

in the GaeSourceRecordBlob

lib/src/dartpad_support_server.dart
```diff
@@ -0,0 +1,122 @@
...
+  @db.StringProperty()
+  String dart;
```
lukechurch: These need to be blobs similar to 
```
  @db.BlobProperty()
  List<int> source;
```

And you'll need to Gzip them, otherwise you'll hit the max length of the strings. See e.g. the FromData in GaeSourceRecordBlob

lib/src/dartpad_support_server.dart
```diff
@@ -0,0 +1,122 @@
...
+class KeyContainer {
+  String key;
+  KeyContainer();
+  KeyContainer.FromKey(String key) {
+    this.key = key;
+  }
+}
+
+// SHA1 set the id
+void sha1SetUUID(GaeExportRecord record) {
```
lukechurch: This name is better thanks.

I think it would be an improvement to have this API compute a SHA that is then set into the object near the callsite.

lukechurch:
@Georgehe4 A few more comments, and please could you add in the api test tool, that'll make it a lot easier to debug. I think with that and these changes we're pretty close to being able to bring this in.

Georgehe4:
@lukechurch 
I have some questions about how the API tests's JSON encoding interacts with the format that the the server receives the data. I will highlight it below.

example/apitest.dart
```diff
@@ -149,6 +175,67 @@ void invokePOST(String api, String source, Element output, {int offset}) {
...
 
+void invokeSupportPOST(String api, support.PadSaveObject pso, Element output, {int offset}) {
+  Stopwatch timer = new Stopwatch()..start();
+  String url = '${_uriBase}${api}';
+  output.text = '';
+
+  Map headers = {'Content-Type': 'application/json; charset=UTF-8'};
+  
+  //Does pso actually have to be a map?
+  String data = JSON.encode(pso); //new Uri(queryParameters: m).query;
```
Georgehe4: @lukechurch 
If I only encode a singular object like this, will the server be able to properly allocated the pso object as parameters of a function

example/apitest.dart
```diff
@@ -6,7 +6,7 @@ library services_server.apitest;
 
 import 'dart:html';
 import 'dart:convert' show JSON;
-
+import '_dartpadsupportservices.dart' as support;
```
Georgehe4: What is the correct way to bring over the PadSupportObject? Should I create some kind of source request instead?

example/apitest.dart
```diff
@@ -149,6 +175,67 @@ void invokePOST(String api, String source, Element output, {int offset}) {
...
+}
+
+void invokeDelete(String api, String key, Element output, {int offset}) {
+  Stopwatch timer = new Stopwatch()..start();
+  String url = '${_uriBase}${api}';
+  output.text = '';
+
+  Map headers = {'Content-Type': 'application/json; charset=UTF-8'};
+
+  String data = JSON.encode(key); //new Uri(queryParameters: m).query;
```
Georgehe4: Same issue; is this the correct way to encode or should I use a map instead?

doc/generated/dartservices.dart
```diff
@@ -2,22 +2,25 @@
 
 library services.dartservices.v1;
 
-import 'dart:async';
+import 'dart:core' as core;
+import 'dart:collection' as collection;
```
Georgehe4: Is collection being used?

lib/services_server.dart
```diff
@@ -122,7 +149,7 @@ class EndpointsServer {
 
     routes = router()
         ..get('/', printUsage)
-        ..add('/api', ['GET', 'POST', 'OPTIONS'], _apiHandler,
+        ..add('/api', ['DELETE', 'GET', 'POST', 'OPTIONS'], _apiHandler,
```
Georgehe4: DELETE no longer needed?

lib/services_gae.dart
```diff
@@ -58,14 +61,15 @@ class GaeServer {
...
+        new rpc.ApiServer(apiPrefix: _API, prettyPrint: true)..addApi(commonServer)
+        ..addApi(fileRelayServer);
   }
 
   Future start() => ae.runAppEngine(requestHandler);
 
   void requestHandler(io.HttpRequest request) {
     request.response.headers.add('Access-Control-Allow-Methods',
-        'POST, OPTIONS');
+        'DELETE, POST, OPTIONS');
```
Georgehe4: Delete no longer needed

doc/generated/dartservices.dart
lukechurch: The auto gen library is a bit pesky like that.

lib/services_server.dart
lukechurch: done.

lukechurch:
@Georgehe4 PTAL - I've redone the rename back to _dartpadsupportservices

tool/grind.dart
```diff
@@ -110,13 +110,24 @@ discovery() {
...
+  discoveryDbFile.writeAsStringSync(resultDb.stdout.trim() + '\n');
+  
   // Generate the Dart library from the json discovery file.
   Pub.global.activate('discoveryapis_generator');
   Pub.global.run('discoveryapis_generator:generate', arguments: [
     'files',
     '--input-dir=doc/generated',
-    '--output-dir=doc/generated',
-    '--no-core-prefixes'
+    '--output-dir=doc/generated'
```
lukechurch: For some reason the '--no-core-reprefixes' option as causing the generator tool to fail now. So I squashed it

Georgehe4:
It looks like we don't address the issue with two users creating the same export at the same time?
Otherwise, lgtm.

lukechurch:
Thanks @Georgehe4 we did address that problem - we're using a 2 part ID: SHA-UUIDv4.

This means that we'd need the people both exporting the same file to have a collision in an RFC4122 v4 UUID (https://www.ietf.org/rfc/rfc4122.txt) to see a problem.

PR #242, opened 2015-07-27
regen the discovery libraries

devoncarew:
Regen the discovery libraries w/ the latest version of the `discoveryapis_generator` package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 build/
 packages
 .buildlog
+.packages
 .pub/
 
 # Or the files created by dart2js.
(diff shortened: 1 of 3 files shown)


devoncarew:
Only whitespace changes -

lukechurch:
But very nice whitespace changes. Is there a change to a dependency we need to continue picking these up?

lgtm.

devoncarew:
The people who regen the libraries just need to ensure that they've run `pub update` locally.

devoncarew:
There may be some min. dep we can set as well...

PR #243, opened 2015-07-29
Summary return types fixed, Server Tests added

Georgehe4:
Return type from AnalysisResult to a String wrapper class (since we cannot return as a String).
@lukechurch

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -268,7 +268,7 @@ class CommonServer {
     if (dart == null || html == null || css == null) {
       throw new BadRequestError('Missing core source parameter.');
     }
-    return await _analyzeMulti({"main.dart" : dart}).then((result) {
+    SummaryText summaryString = await _analyzeMulti({"main.dart" : dart}).then((result) {
       Summarizer summarizer = new Summarizer(
           dart: dart,
           html: html,
@@ -276,6 +276,7 @@ class CommonServer {
           analysis: result);
       return new SummaryText.fromString(summarizer.returnAsSimpleSummary());
     });
+    return new Future.value(summaryString);
   }
 
   Future<AnalysisResults> _analyzeMulti(Map<String, String> sources) async {
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -44,6 +44,11 @@ abstract class SourceRequestRecorder {
   Future record(String verb, String source, [int offset]);
 }
 
+class SummaryText {
+  String text;
+  SummaryText.fromString(String this.text);
+}
+
 abstract class PersistentCounter {
   Future increment(String name, {int increment: 1});
   Future<int> getTotal(String name);
@@ -114,7 +119,7 @@ class CommonServer {
       path: 'summarize',
       description: 'Summarize the given Dart source code and return any resulting '
         'analysis errors or warnings.')
-  Future<AnalysisResults> summarize(SourcesRequest request) {
+  Future<SummaryText> summarize(SourcesRequest request) {
     return _summarize(request.sources['dart'], request.sources['css'], request.sources['html']);
   }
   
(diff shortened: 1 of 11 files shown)


lukechurch:
@Georgehe4 This needs a test before we merge.

Georgehe4:
Pushing this back until the week is done.

Georgehe4:
@devoncarew @lukechurch 
Ready for review.

lib/src/dartpad_support_server.dart
```diff
@@ -19,17 +20,75 @@ final Logger _logger = new Logger('dartpad_support_server');
...
+  var database;
+  bool test;
+
+  String getTypeName(dynamic obj) =>
+      mirrors.reflect(obj).type.reflectedType.toString();
+
+  String getClass(obj) =>
+      mirrors.MirrorSystem.getName(mirrors.reflectClass(obj).simpleName);
+
+  FileRelayServer({bool test}) {
```
devoncarew: `bool test: false`

lib/src/dartpad_support_server.dart
```diff
@@ -19,17 +20,75 @@ final Logger _logger = new Logger('dartpad_support_server');
...
     _logger.level = Level.ALL;
+    if (test) {
+      database = new Map();
+      this.test = true;
+    } else {
+      this.test = false;
+    }
+  }
+
+  Future<List> databaseQuery(var dbClass, String attribute, var value) async {
```
devoncarew: If these methods are only used in this class lets make them private (here and `databaseCommit`)

lib/src/dartpad_support_server.dart
```diff
@@ -19,17 +20,75 @@ final Logger _logger = new Logger('dartpad_support_server');
...
+  String getTypeName(dynamic obj) =>
+      mirrors.reflect(obj).type.reflectedType.toString();
+
+  String getClass(obj) =>
+      mirrors.MirrorSystem.getName(mirrors.reflectClass(obj).simpleName);
+
+  FileRelayServer({bool test}) {
     hierarchicalLoggingEnabled = true;
     _logger.level = Level.ALL;
+    if (test) {
```
devoncarew: Using an `if` statements here totally works. Another way to do it is for the `ae.context.services.db.query` functionality to be injected in via the constructor. So, you have an interface which just exposes the operations you use. You have one implementation which implements that interface using the appengine class. It'll pretty much be a 1:1 passthrough for the calls you use. You then have a 2nd implementation which is a mock one. The real impl is the one that this class defaults to if none is provided.

test/dartpad_server_test.dart
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
```
devoncarew: 2015

devoncarew:
I opened a bug for the test timeouts.

lgtm! @lukechurch?

lib/src/common_server.dart
```diff
@@ -259,17 +264,17 @@ class CommonServer {
...
   }
   
-  Future<AnalysisResults> _summarize(String dart, String html, String css) async {
+  Future<SummaryText> _summarize(String dart, String html, String css) async {
     if (dart == null || html == null || css == null) {
       throw new BadRequestError('Missing core source parameter.');
     }
-    return _analyzeMulti({"main.dart" : dart}).then((result) {
-      Summarizer summer = new Summarizer(
+    return await _analyzeMulti({"main.dart" : dart}).then((result) {
```
lukechurch: I don't really like the return await, it's confusing as to the relationship between await, and then.

Let refactor this to the more standard await foo.
return bar;

lib/src/dartpad_support_server.dart
```diff
@@ -59,16 +120,14 @@ class FileRelayServer {
...
     List result;
     do {
       randomUuid = new uuid_tools.Uuid().v4();
-      query = database.query(_GistMapping)..filter('internalId =', randomUuid);
-      result = await query.run().toList();
-      attemptCount ++;
-      if (!result.isEmpty) _logger.info("Collision in retrieving mapping id ${randomUuid}.");
+      result = await _databaseQuery(_GistMapping, 'internalId =', randomUuid);
+      attemptCount++;
+      if (!result.isEmpty) _logger
```
lukechurch: This is awkward formatting please change to if (...) {
  _logger.info(...)
}

lukechurch:
lgtm

Georgehe4:
Merging.

PR #245, opened 2015-08-06
Backend for gist rerouting

Georgehe4:
Adding mappings to enable link-back to dartpad from github gists.
Ready for review.


lukechurch:
@Georgehe4 Is this ready for review other than the tests?

Georgehe4:
Not yet. I'll ping you and Devon when ready.

Georgehe4:
@devoncarew @lukechurch 
Ready for review.

lib/src/dartpad_support_server.dart
```diff
@@ -94,6 +159,18 @@ class UuidContainer {
 }
 
 /**
+ * Map from id to id
+ */
+class Mapping {
+  String gistId;
+  String internalId;
+  Mapping();
+  Mapping.FromIds(String gistId, String internalId) {
```
devoncarew: Let's change this to `Mapping.fromIds(this.gitsIs, this.internalId);`

lib/src/dartpad_support_server.dart
```diff
@@ -58,6 +58,68 @@ class FileRelayServer {
     _logger.info("Deleted Export with ID ${record.uuid}");
     return new Future.value(new PadSaveObject.FromRecordSource(record));
   }
+
+  @ApiMethod(method: 'GET', path: 'getValidId')
+  Future<UuidContainer> getValidId() async {
+    print('getValidId');
```
devoncarew: Here and a few other places, debugging `print`s.

Georgehe4: @lukechurch There are some print statements in the previous support services functions landed a while back. Should we remove those as well?

lib/src/dartpad_support_server.dart
```diff
@@ -39,7 +39,7 @@ class FileRelayServer {
     return new Future.value(new UuidContainer.FromUuid(record.uuid));
   }
 
-  @ApiMethod(method: 'POST', path: 'pullExportData')
+  @ApiMethod(method: 'POST', path: 'pullExportData', description: 'Retrieve a stored gist data set.')
   Future<PadSaveObject> pullExportContent(UuidContainer uuidContainer) async {
     print ('pullExportContent');
```
Georgehe4: @lukechurch Print statements such as these.

Georgehe4:
@devoncarew @lukechurch Changes made.

devoncarew:
lgtm for my part

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
       throw (e);
     });
     _logger.info("Deleted Export with ID ${record.uuid}");
-    return new Future.value(new PadSaveObject.FromRecordSource(record));
+    return new Future.value(new PadSaveObject.fromRecordSource(record));
+  }
+
+  @ApiMethod(method: 'GET', path: 'getValidId')
+  Future<UuidContainer> getValidId() async {
```
lukechurch: This doesn't really work as a method name. 'valid' is superfluous -- why would we want an invalid id? That leaves 'getId' -- what's it an Id for?

Is this something like getGistMappingId?

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
...
+    int count = 0;
+    int limit = 4;
```
lukechurch: These should be final constants

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
...
+    List result;
+    do {
+      randomUuid = new uuid_tools.Uuid().v4();
+      query = database.query(_GistMapping)..filter('internalId =', randomUuid);
+      result = await query.run().toList();
+      count ++;
+    } while (!result.isEmpty && count < limit);
+    if (!result.isEmpty) {
+      _logger.severe("Could not generate valid ID.");
+      return new Future.value(new UuidContainer());
```
lukechurch: Please throw an exception here instead. We need to give an Err 500 back to the user

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
       throw (e);
     });
     _logger.info("Deleted Export with ID ${record.uuid}");
-    return new Future.value(new PadSaveObject.FromRecordSource(record));
+    return new Future.value(new PadSaveObject.fromRecordSource(record));
+  }
+
+  @ApiMethod(method: 'GET', path: 'getValidId')
+  Future<UuidContainer> getValidId() async {
+    int count = 0;
```
lukechurch: attemptCount?

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
...
+  }
+
+  @ApiMethod(method: 'POST', path: 'storeGist')
+  Future<UuidContainer> storeGist(Mapping map) async {
+    var database = ae.context.services.db;
+    var query = database.query(_GistMapping)..filter('internalId =', map.internalId);
+    List result = await query.run().toList();
+    if (!result.isEmpty) {
+      _logger.severe("Collision with mapping of Id ${map.gistId}.");
+      return new Future.value(new UuidContainer());
```
lukechurch: Throw an exception here instead

lib/src/dartpad_support_server.dart
lukechurch: maybe "getUnusedMappingId"?

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
...
+    int limit = 4;
+    var database = ae.context.services.db;
+    String randomUuid;
+    var query;
+    List result;
+    do {
+      randomUuid = new uuid_tools.Uuid().v4();
+      query = database.query(_GistMapping)..filter('internalId =', randomUuid);
+      result = await query.run().toList();
+      count ++;
```
lukechurch: The probability of collisions here is so low that I think it would be a good idea to log anytime we see a collision.

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
...
+  }
+
+  @ApiMethod(method: 'POST', path: 'retrieveGist')
+  Future<UuidContainer> retrieveGist(UuidContainer id) async {
+    var database = ae.context.services.db;
+    var query = database.query(_GistMapping)..filter('internalId =', id);
+    List result = await query.run().toList();
+    if (result.isEmpty) {
+      _logger.severe("Missing mapping for Id ${id.uuid}.");
+      return new Future.value(new UuidContainer.fromUuid(""));
```
lukechurch: This should return an Error 400

lib/src/dartpad_support_server.dart
```diff
@@ -56,7 +52,63 @@ class FileRelayServer {
...
+      db.dbService.commit(inserts: [entry]).catchError((e) {
+        _logger.severe("Error while recording mapping with Id ${map.gistId}. Error ${e}");
+        throw e;
+      });
+      _logger.info("Mapping with ID ${map.gistId} stored.");
+      return new Future.value(new UuidContainer.fromUuid(map.gistId));
+    }
+  }
+
+  @ApiMethod(method: 'POST', path: 'retrieveGist')
```
lukechurch: I think this should be a GET method not a POST?

lib/src/dartpad_support_server.dart
```diff
@@ -85,15 +137,30 @@ class PadSaveObject {
...
-  UuidContainer.FromUuid(String uuid) {
+  UuidContainer.fromUuid(String uuid) {
     this.uuid = uuid;
   }
 }
 
 /**
+ * Map from id to id
+ */
+class Mapping {
```
lukechurch: Lets be more specific here, it's a GistToInternalIdMapping?

lukechurch:
Thanks for this @Georgehe4 there's a bit to do on fixing up the handling of error states within the server and being more specific in some of the method names. In general the server shouldn't try and silently absorb cases where it has a problem - e.g. missing data, it should use an HTTP error code to tell the client that either the data it was sent was wrong (err 4xx series), or that it experienced an internal problem (err 5xx series)

Georgehe4:
@lukechurch Let me know if any of the errors or names still need fixing.

lukechurch:
Thanks @Georgehe4 lgtm.

PR #249, opened 2015-08-31
add more version info

devoncarew:
Return the full version of the dartpad backend (as well as the version summary we currently have). This will let the frontend display `1.12.0` _and_ the additionally detail of `1.12.0+2`, which can be useful when diagnosing some issues.


lukechurch:
lgtm

PR #250, opened 2015-08-31
1.12.0

devoncarew:
Rev the compiler package to the latest stable sdk (`1.12.0`).

@lukechurch


lukechurch:
@devoncarew Thanks, - I'll also need to change the docker file etc. I'll ammend the PR a bit before merging

PR #251, opened 2015-09-15
compiler 1.12.1

devoncarew:
@lukechurch


devoncarew:
Landing, this changes the version of dart2js used, but 1.12.1 does not actually have any dart2js changes.

Issue #253, opened 2015-10-14
new compiler_unsupported is available


devoncarew:
`1.13.0-dev.7.0` - this is very close to what `1.13.0` stable will be

@lukechurch

devoncarew:
ping! `1.13.0` is imminent - let me know if I can help out

lukechurch:
Thanks @devoncarew

Issue #247, opened 2015-08-17
build is failing on analysis test failures


devoncarew:
See https://travis-ci.org/dart-lang/dart-services/builds/75686478 and https://github.com/dart-lang/dart-services/pull/243.

devoncarew:
fixed

Issue #237, opened 2015-07-08
use the analysis server binary from the SDK?


devoncarew:
Should we switch over to using the analysis server binary from the SDK, or is it still important to track the source from HEAD?

lukechurch:
Unfortunately we can't do this until at least the new task model having landed.

lukechurch:
Fixed in https://github.com/dart-lang/dart-services/pull/257

PR #255, opened 2016-01-03
fix regressions; upgrade to 1.13.1

devoncarew:
Upgrade to `1.13.1`.

- upgrade to `compiler_unsupported` 1.13.1
- fix a regression where we threw an NPE if Platform.packageRoot was null
- fix ~30 unit test failures due to content type handling
- fix an issue with changed compilation errors from dart2js

@lukechurch 

Luke, can you test this w/ appengine when it lands? The local server is working. There were a lot of various changes and fixes to fix regressions, and then to upgrade to the latest packages.

We'll also need to rev from analysis server 1.12.0 to something more recent.

### DIFF
--- a/.atom/launches/services.yaml
+++ b/.atom/launches/services.yaml
@@ -1,0 +1,8 @@
+# Cli launch configuration for bin/services.dart.
+type: cli
+path: bin/services.dart
+
+cli:
+  args: 
+  checked: true
+  debug: true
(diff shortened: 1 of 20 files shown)


devoncarew:
fix https://github.com/dart-lang/dart-services/issues/253, fix https://github.com/dart-lang/dart-services/issues/247

devoncarew:
Re: upgrading the analysis server, we should probably implement https://github.com/dart-lang/dart-services/issues/237.

lukechurch:
@devoncarew I'll begin testing this on Monday, there are a lot of revisions all over the place, including in GAE deploy infrastructure.

PR #257, opened 2016-01-12
Upgrade dart-services to 1.13

lukechurch:
- Defork Analysis Server and launch from the SDK, inlining the protocol
- Adjust launch process to support dev_appserver to allow local testing

Addresses #253 #237


lukechurch:
Build failure is caused by:   

[info] 'computeDocumentationComment' is deprecated at lib/src/analyzer.dart, line 256.

I'll address in a follow on commit

PR #261, opened 2016-01-12
Force interpretation of text as JSON

lukechurch:
Dartpad sends data as text/plain to avoid an extra OPTIONS round trip.

This needs to be promoted before it's passed to the rpc library in order for it to dispatch correctly.

### DIFF
--- a/tool/warmup.dart
+++ b/tool/warmup.dart
@@ -48,7 +48,7 @@ request(String verb, String postPayload) async {
   return http.post(
       Uri.parse(uri + verb),
       body: postPayload,
-      headers: {'content-type': 'application/json; charset=utf-8'}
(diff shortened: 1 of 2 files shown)


tool/warmup.dart
```diff
@@ -48,7 +48,7 @@ request(String verb, String postPayload) async {
   return http.post(
       Uri.parse(uri + verb),
       body: postPayload,
-      headers: {'content-type': 'application/json; charset=utf-8'}
+       headers: {'content-type': 'text/plain; charset=utf-8'}
```
devoncarew: nit: extra space here

lukechurch: thanks

Issue #269, opened 2016-04-15
CryptoUtils is deprecated


lukechurch:
Need to point the crypto methods to use the new structuring

lukechurch:
Resolved some time ago

PR #270, opened 2016-04-15
Support strong mode

lukechurch:
Upgrade to the latest version of the VM, add an optional argument to support strong mode.

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -62,7 +62,9 @@ class CommonServer {
   final PersistentCounter counter;
 
   Pub pub;
+  Analyzer strongModeAnalyzer;
   Analyzer analyzer;
+
   Compiler compiler;
   AnalysisServerWrapper analysisServer;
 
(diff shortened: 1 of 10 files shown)


lukechurch:
Build issues are cosmetic, I'll get a fix shortly: https://github.com/dart-lang/dart-services/issues/269

lib/src/common_server.dart
```diff
@@ -275,24 +279,28 @@ class CommonServer {
...
       throw new BadRequestError('Missing parameter: \'sources\'');
     }
     Stopwatch watch = new Stopwatch()..start();
     String sourcesJson = new JsonEncoder().convert(sources);
-    srcRequestRecorder.record("ANALYZE-v1", sourcesJson);
+    srcRequestRecorder.record("ANALYZE-v2-$strongMode", sourcesJson);
     _logger.info("About to ANALYZE-v1: ${_hashSource(sourcesJson)}");
+
+    // Select the right analyzer
+    Analyzer selectedAnalyzer = strongMode ? strongModeAnalyzer : analyzer;
```
devoncarew: The `strongMode` param is coming from the request - you may want to guard against it being null here. Something like `strongMode ??= false` before this line.

lukechurch: It'll default to false if it is null at the request handler, but I've no objection to putting in the extra-defensiveness.

devoncarew: I'm you're sure, then there's no need. I was worried tracing through the logic that we could be passing a `null` value into a named param, which would override the default of `strongMode: false`.

lukechurch: I prefer the check - added, thanks for the catch.

PR #272, opened 2016-04-16
upgrade to 1.16.0

devoncarew:
- upgrade to 1.16.0

@lukechurch


lukechurch:
Thanks @devoncarew, I'll pull this in and run a more extensive test. I'm going to let things settle for a week or so after the strong mode push. Assuming all is good, I'll do a push of this version in a week or so.

Thanks for upgrading the compiler.

PR #275, opened 2016-11-17
Remove librato reporting benchmarks

lukechurch:
@devoncarew TBR

The librato integration is causing a version lock that is preventing me rolling analyzer forwards. I'll need to remove it. If you feel strongly about keeping it we can add it back in again later once we've addressed the version issue


lukechurch:
I'm aware of the warnings about depreciations that's causing Travis to complain, I'll be fixing those as part of this arc of work.

devoncarew:
lgtm; no issues w/ the removal :)

PR #276, opened 2016-11-17
Upgrade crypto lib

lukechurch:
@devoncarew @danrubel 

We need to upgrade this lib so that we can upgrade analyzer.

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -293,7 +293,7 @@ class CommonServer {
       throw new BadRequestError('Missing parameter: \'sources\'');
     }
     strongMode ??= false;
-    
+
     Stopwatch watch = new Stopwatch()..start();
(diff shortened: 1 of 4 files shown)


devoncarew:
lgtm

lib/src/common_server.dart
```diff
@@ -545,7 +545,5 @@ String _printCompileProblem(CompilationProblem problem) {
 }
 
 String _hashSource(String str) {
-  SHA1 sha1 = new SHA1();
-  sha1.add(str.codeUnits);
-  return CryptoUtils.bytesToHex(sha1.close());
+  sha1.convert(str.codeUnits).toString();
```
lukechurch: Yes. The crypto library now surfaces the algorithms as top level fields, and makes the constructors private.

PR #285, opened 2017-02-22
Add strong mode to fuzz tester

lukechurch:
@danrubel


danrubel:
LGTM. We need to address the strong mode nit in another PR.

lukechurch:
Thanks, and yes, I agree.

PR #287, opened 2017-03-21
Upgrade VM configuration to use flex environments

lukechurch:
FYI @danrubel @devoncarew 

This moves dart-services into the new vm configuration from for appengine.


lukechurch:
I'm addressing the failures that Travis highlighted

danrubel:
LGTM once Travis issues addressed.

lukechurch:
The remaining problem is an info message about an unnecessary cast in code that we're importing from analysis server. I'm going to merge this now and fix the cast in a seperate PR as it'll need more extensive testing.

PR #290, opened 2017-04-10
Analyzer upgrade dan

danrubel:
A few more fixes for analyzer integration in DartPad

### DIFF
--- a/lib/src/analyzer.dart
+++ b/lib/src/analyzer.dart
@@ -72,9 +72,11 @@ class Analyzer {
     PhysicalResourceProvider physicalResourceProvider =
               PhysicalResourceProvider.INSTANCE;
 
-    AnalysisOptionsImpl options = new AnalysisOptionsImpl();
-    options.enableGenericMethods = true;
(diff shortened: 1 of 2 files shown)


lib/src/analyzer.dart
```diff
@@ -72,9 +72,11 @@ class Analyzer {
     PhysicalResourceProvider physicalResourceProvider =
               PhysicalResourceProvider.INSTANCE;
 
-    AnalysisOptionsImpl options = new AnalysisOptionsImpl();
-    options.enableGenericMethods = true;
-    options.strongMode = strongMode;
+    AnalysisOptionsImpl analysisOptions = new AnalysisOptionsImpl();
+    analysisOptions.strongMode = strongMode;
```
lukechurch: I think we need the flag to enable Generic methods? We certainly used to need this - we had users complain about the behaviour of strong mode otherwise

danrubel: Generic methods are enabled by default in HEAD.

PR #291, opened 2017-04-10
use absolute paths when calling analyzer

danrubel:
* use absolute paths
* update changed content to include source
* update tests


lukechurch:
lgtm!

PR #292, opened 2017-04-19
remove implicit sources

danrubel:
When a source contains an import of a file/source that is undefined, then analyzer implicitly creates a source representing that non-existing source. This change removes that implicitly created source after one analysis request so that it does not interfere with subsequent analysis requests.


lukechurch:
LGTM, thanks!

PR #299, opened 2017-04-28
dartfmt

devoncarew:
@lukechurch, dartfmt all the sources - should make future PRs a bit easier


lukechurch:
Great, thanks.

LGTM

Issue #256, opened 2016-01-03
look into using `test`


devoncarew:
We're using `unittest` right now, which is fine, but it's not being maintained, and we're getting older versions of packages pulled in via transitive deps. At some point we'll run into version constraint issues when we do need to use a newer version of one of the transitive packages.

devoncarew:
Looks like `shelf_route` is pinning us to older packages, incompatible with the current `package:test`.

PR #301, opened 2017-04-29
use package:test to run tests

devoncarew:
- switch to using package:test to run tests. This is largely a mechanical change, w/ some work to get the pubspec right - this will get simpler if and when we drop the dep on 'package:analyzer`
- fix https://github.com/dart-lang/dart-services/issues/256

@danrubel @lukechurch


pubspec.yaml
```diff
@@ -22,15 +22,20 @@ dependencies:
...
-  unittest: ^0.11.0
+  test: ^0.12.0
 
 executables:
   services: null
+
+dependency_overrides:
+  # package:test's version constraint does not include analyzer 0.30.0 alpha
+  analyzer: 0.30.0-alpha.3
+  # package:shelf_route's version constraint doesn't not include matcher 0.12.0
```
lukechurch: nit 'doesn't not' -> 'does not'?

lukechurch:
LGTM with nit

When we're done with this, lets also chat about how we might integrate the longer tests that we run as a pre-deployment script into this.

pubspec.yaml
devoncarew: oops

devoncarew:
Yeah, there was some legerdemain around `gae_deployed_test.dart` - to keep the spirit of the `unittest` based tests - but definitely something we can improve here.

Issue #302, opened 2017-05-01
re-evaluate our package dependencies


devoncarew:
We're using some older, less well maintained packages in dart-services. These cause us to select older versions of other packages, and will get us into trouble going forward (we won't be able to rev dartpad without forcing the use of current package versions in the pubspec). We should re-evaluate the set of packages we're depending on.

Issue #298, opened 2017-04-27
explore a lighter-weight library for communicating with the analysis server


devoncarew:
This issue is not critical, but I do have a lightweight library - in use by Atom and tuenup - for communicating with the analysis server. Switching to this library could reduce and simplify the amount of code in dart-services.

lukechurch:
+1.

I suggest we actually look at doing this before moving from analyzer to analysis server. I'm happy to help out in the process.

PR #304, opened 2017-05-10
Refactor analysis server usage

devoncarew:
Some work towards https://github.com/dart-lang/dart-services/issues/298 (and https://github.com/dart-lang/dart-services/issues/302):

- remove the current analysis server communications library (`lib/src/analysis_server_protocol/`)
- bring a new one in via pubspec imports (`package:analysis_server_lib`)
- update `lib/src/analysis_server.dart` to use the new library
- fix tests associated with the change (mostly, updating some formatter service tests)
- re-enable some analysis server benchmarks
- package rename from `services` to `dart_services`, mostly to make stack traces that include package refs from this package a bit easier to read
- removed a `package:shelf_route` dep, only used in the non-gae (testing) version of the server. this package's deps were not updated, and were causing us to select old package versions
- removed use of `package:cli_util` to locate the sdk - we have more accurate APIs available in dart now

Tests run and pass; @lukechurch, @danrubel please to review :)

@lukechurch, this is a bigger change (though in-line w/ what we've discussed). I'd like to keep building on it towards the direction of removing our dep on `package:analyzer`. I can branch from this PR if you'd like more time to review, or to discuss when you're in town next week.


devoncarew:
And for fun, the benchmark results:

```
Running 9 benchmarks.

[analyzer.hello      :    1.925ms]
[analyzer.hellohtml  :    2.005ms]
[analyzer.sunflower  :    4.775ms]
[completion.hello    :    4.223ms]
[completion.hellohtml:    5.771ms]
[completion.sunflower:   11.888ms]
[dart2js.hello       :  437.614ms]
[dart2js.hellohtml   : 1411.960ms]
[dart2js.sunflower   : 2360.742ms]
```

danrubel:
LGTM, but I'd like @lukechurch input on this

lib/src/common_server.dart
```diff
@@ -502,24 +504,14 @@ class CommonServer {
...
     }
-    if (offset == null) offset = 0;
+    offset ??= 0;
+
     Stopwatch watch = new Stopwatch()..start();
     srcRequestRecorder.record("FORMAT", source, offset);
     _logger.info("About to FORMAT: ${_hashSource(source)}");
     counter.increment("Formats");
 
-    // Guard against trying to format code with errors.
```
lukechurch: The idea for removing this is that the analysis server will just hand the same source back if we request a format on code with errors?

lib/src/analysis_server.dart
```diff
@@ -217,413 +249,82 @@ class AnalysisServerWrapper {
...
-      {Map<String, String> packageRoots}) {
-    var params = new AnalysisSetAnalysisRootsParams(included, excluded,
-            packageRoots: packageRoots)
-        .toJson();
-    return send("analysis.setAnalysisRoots", params);
+    return analysisServer.analysis.updateContent(params);
   }
 
-  Future dispatchNotification(String event, params) async {
-    if (event == "server.error") {
```
lukechurch: I couldn't see (possibly just because of the diff UI) where the equivalent error handling code is? I'd like to be extra careful looking at that, as it's important that we don't ever allow failed servers to start queueing requests.

lib/src/common_server.dart
devoncarew: Yup. The analysis server call handles code that doesn't parse already - it will return an error (code that parses but isn't semantically correct is still useful to format). In the wrapepr around the analysis server, we convert those errors to returning a format response w/ exactly the code that was passed in. That was done in order to have the least impact on the front-end. Going forward though, it's probably useful to be able to discriminate between 'no formatting changes' and 'there were problems formatting'.

lib/src/analysis_server.dart
devoncarew: Ah! The PR doesn't have the same restart-on-error logic but I can add such.

devoncarew:
Thanks for the reviews! Will update the PR w/ some logic to restart the server on analysis server errors. (the good news is that we've dropped the failure rate of the analysis server by an order of magnitude)

lib/src/analysis_server.dart
devoncarew: Updated

devoncarew:
Re:

> I couldn't see (possibly just because of the diff UI) where the equivalent error handling code is? I'd like to be extra careful looking at that, as it's important that we don't ever allow failed servers to start queueing requests.

I updated the PR to listen for errors from the analysis server and log them to the logger (`_logger.severe(...)`). I also added logic to listen for analysis server process termination and exit the serving process w/ the same exit code.

I don't think we want to put in place logic to restart the analysis server when it reports errors. We had such in IntelliJ and recently removed it. The errors reported by the AS are generally just related to the request and don't leave the AS in a place where it can't respond to future requests.

We can put in place different or more strict hardening logic, but I think it would be worth a discussion first :)

Issue #297, opened 2017-04-27
refactor to not use package:analyzer


devoncarew:
It makes rolls more difficult - there's often not a package:analyzer to corresponds to the version of the SDK we want to target. I think we can use the analysis server from the SDK directly.

PR #305, opened 2017-05-14
change the document and analyze services to use the analysis server

devoncarew:
- change the document and analyze services to use the analysis server (work towards https://github.com/dart-lang/dart-services/issues/297)
- rename `bin/services.dart` to `bin/server_dev.dart` to make it clear that it's not the prod server
- upgrade to the latest `analysis_server_lib` lib version
- fix a race condition with getFixes, completions, and document (found using the fuzz tester)
- update the benchmarks and fuzz tester to use the analysis server based services

@lukechurch, @danrubel - things are looking good - dartpad (client and server) work locally, tests pass, and the fuzz tester seems happy.

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -10,9 +10,11 @@ import 'dart:io';
 
 import 'package:analysis_server_lib/analysis_server_lib.dart';
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as path;
 
 import 'api_classes.dart' as api;
 import 'common.dart';
+import 'pub.dart';
 import 'scheduler.dart';
 
 final Logger _logger = new Logger('analysis_server');
@@ -40,10 +42,13 @@ class AnalysisServerWrapper {
   AnalysisServerWrapper(this.sdkPath) {
     _logger.info("AnalysisServerWrapper ctor");
     sourceDirectory = Directory.systemTemp.createTempSync('analysisServer');
-    // TODO: We should write an analysis_options.yaml file here with strong mode
-    // enabled.
     mainPath = _getPathFromName(kMainDart);
+
     serverScheduler = new TaskScheduler();
+
+    // Write an analysis_options.yaml file with strong mode enabled.
+    File optionsFile = new File(_getPathFromName('analysis_options.yaml'));
+    optionsFile.writeAsStringSync('analyzer:\n  strong-mode: true\n');
   }
 
   Future init() {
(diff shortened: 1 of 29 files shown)


lib/src/analysis_server.dart
```diff
@@ -37,11 +39,16 @@ class AnalysisServerWrapper {
   /// Instance to handle communication with the server.
   AnalysisServer analysisServer;
 
-  AnalysisServerWrapper(this.sdkPath) {
+  AnalysisServerWrapper(this.sdkPath, bool useStrongMode) {
     _logger.info("AnalysisServerWrapper ctor");
```
lukechurch: Could you stick the value of useStrongMode in the logging call?

_logger.info("AnalysisServerWrapper ctor, strong mode: $useStrongMode");

tool/fuzz_driver.dart
```diff
@@ -135,7 +135,7 @@ setupTools(String sdkPath) async {
   apiServer = new ApiServer(apiPrefix: '/api', prettyPrint: true)
     ..addApi(server);
 
-  analysisServer = new analysis_server.AnalysisServerWrapper(sdkPath);
+  analysisServer = new analysis_server.AnalysisServerWrapper(sdkPath, true);
```
lukechurch: We need to drive both the strong and non-Strong analysis server with these calls. I can fix this up, could you add a TODO for me?

tool/fuzz_driver.dart
```diff
@@ -432,8 +431,14 @@ enum OperationType {
...
-log(dynamic str) {
+final int termWidth = io.stdout.hasTerminal ? io.stdout.terminalColumns : 200;
+
+void log(dynamic obj) {
   if (_VERBOSE) {
-    print("${new DateTime.now()} $str");
+    String str = '${new DateTime.now()} $obj';
+    if (str.length > termWidth) {
+      str = str.substring(0, termWidth - 4) + '...';
+    }
```
lukechurch: Let's not do this in the case of fuzz_driver, we parse the results of these files with support tools for downstream analysis.

lukechurch: If you wish, we can change that in a subsequent PR, but for now, let's just output the whole thing.

devoncarew: sure, reverted

tool/fuzz_driver.dart
devoncarew: makes sense; I'll add a todo

lib/src/analysis_server.dart
```diff
@@ -37,11 +39,16 @@ class AnalysisServerWrapper {
   /// Instance to handle communication with the server.
   AnalysisServer analysisServer;
 
-  AnalysisServerWrapper(this.sdkPath) {
+  AnalysisServerWrapper(this.sdkPath, bool useStrongMode) {
```
devoncarew: done

PR #306, opened 2017-05-18
remove analyzer libraries

devoncarew:
- remove the last use of analyzer libraries
- fix https://github.com/dart-lang/dart-services/issues/297

@lukechurch


lukechurch:
LGTM, thanks!

PR #309, opened 2017-06-12
Pub package url update

lukechurch:
TBR @devoncarew The base URL for pub has changed which was causing some of the tests to fail. From the pubsite this seems to be the url in current use.

### DIFF
--- a/lib/src/pub.dart
+++ b/lib/src/pub.dart
@@ -179,7 +179,7 @@ class Pub {
   Future _populatePackage(
       PackageInfo package, Directory cacheDir, Directory target) {
     final String base =
-        'https://storage.googleapis.com/pub.dartlang.org/packages';
+        'https://storage.googleapis.com/pub-packages/packages';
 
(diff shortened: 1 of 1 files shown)


lukechurch:
cc @danrubel

devoncarew:
lgtm (and /cc @mkustermann)

lib/src/pub.dart
```diff
@@ -179,7 +179,7 @@ class Pub {
   Future _populatePackage(
       PackageInfo package, Directory cacheDir, Directory target) {
     final String base =
-        'https://storage.googleapis.com/pub.dartlang.org/packages';
+        'https://storage.googleapis.com/pub-packages/packages';
```
mkustermann: It is, generally speaking, not safe to rely on these urls. As you discovered, we changed the GCS bucket we use to store the tarballs.

You could use the official api to get the package:

    https://pub.dartlang.org/packages/<name>/versions/<version>.tar.gz

The http response will redirect you to the correct place.

PR #311, opened 2017-06-12
Add test corpus from gists

lukechurch:
@devoncarew 

As discussed with @danrubel it would be useful to have a corpus of public dart files that we can run the fuzz tester on ahead of each release. This adds a corpus based on the sample gists.

### DIFF
--- a/test/corpus/bootstrap_gist.dart
+++ b/test/corpus/bootstrap_gist.dart
@@ -1,0 +1,15 @@
+// Copyright 2015 the Dart project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license
+// that can be found in the LICENSE file.
+
+import 'dart:html';
+
+void main() {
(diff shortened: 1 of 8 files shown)


danrubel:
LGTM

devoncarew:
Re: the failures (https://travis-ci.org/dart-lang/dart-services/builds/242044827#L389), I'd suggest adding an excludes line in the analysis options to exlcude the corpus directory from analysis.

+1 to the corpus - sounds great. These files are going to get out of sync with the samples we ship however. I'd recommend adding a grind task to git clone the (10?) samples we ship with and move their dart source into the corpus. We can run this script manually - it'll be much easier to keep the corpus up to date.

lgtm w/ the analysis fix

devoncarew:
(I did update those two samples to declare return types)

lukechurch:
Thanks @devoncarew, I'll follow the grind task approach

Issue #317, opened 2017-08-16
look into being able to specify the version of the sdk that dart-services should use


devoncarew:
Something like having a version file, and on boostrap, dart-services will download that sdk version if it's not already present locally.

PR #318, opened 2017-08-17
decouple the dart sdk version from the host sdk version

devoncarew:
@danrubel @lukechurch 

decouple the dart sdk version from the host sdk version; fix https://github.com/dart-lang/dart-services/issues/317

This introduces a top-level file in the repo, `dart-sdk.version`. When the dart-services server starts, it checks for a dak sdk in the `dart-sdk/` directory, and does a version check against the version in that `dart-sdk.version`. If the sdk does not exist, or the version is different, the serve will download a new version of the sdk (for macos or linux), uncompress it, and place it in `dart-sdk`. (this is actually fairly similar to the flutter sdk upgrade code).

This will allow us to more easily rev dartpad to track sdk releases.


devoncarew:
This PR also revs dart-services to the very latest dev sdk release. That's not a critical component of the PR - we can easily switch to whatever version is desired.

lib/src/sdk_manager.dart
```diff
@@ -0,0 +1,91 @@
...
+
+    print('Downloading Dart SDK $version...');
+    ProcessResult result = await Process.run('curl',
+        ['-continue-at=-', '--location', '--output', destFile.path, url]);
+    if (result.exitCode != 0) {
+      throw 'curl failed: ${result.exitCode}\n${result.stdout}\n${result
+          .stderr}';
+    }
+    result = await Process
+        .run('unzip', ['-o', '-q', destFile.path, '-d', path.dirname(sdkPath)]);
```
danrubel: Should the unzip destination directory be created if it does not exist? Deleted and recreated so that there are no conflicts with a older version that was downloaded previously?

devoncarew: It overwrite correctly (but I will verify locally). The unzip directory is currently the cwd (the root of the project). The zip contains a directory (`dart-sdk`), which is what gives us the sub-dir in the project root.

I'll add a create if it doesn't exist step.

lib/src/sdk_manager.dart
```diff
@@ -0,0 +1,96 @@
...
+      throw 'curl failed: ${result.exitCode}\n${result.stdout}\n${result
+          .stderr}';
+    }
+
+    Directory destDir = new Directory(path.dirname(sdkPath));
+    if (!destDir.existsSync()) {
+      destDir.createSync(recursive: true);
+    }
+    result = await Process
+        .run('unzip', ['-o', '-q', destFile.path, '-d', destDir.path]);
```
lukechurch: unzip isn't available by default on the gae instances. Would gunzip work?

lib/src/sdk_manager.dart
```diff
@@ -0,0 +1,96 @@
...
+    if (result.exitCode != 0) {
+      throw 'unzip failed: ${result.exitCode}\n${result.stdout}\n${result
+          .stderr}';
+    }
+    print('SDK available at $sdkPath');
```
lukechurch: This and other places using print should probably use the logging package to make sure the data gets routed to the right place on the server.

lib/src/sdk_manager.dart
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    Directory destDir = new Directory(path.dirname(sdkPath));
+    if (!destDir.existsSync()) {
+      destDir.createSync(recursive: true);
+    }
+    result = await Process
+        .run('unzip', ['-o', '-q', destFile.path, '-d', destDir.path]);
+    if (result.exitCode != 0) {
+      throw 'unzip failed: ${result.exitCode}\n${result.stdout}\n${result
```
lukechurch: What do we do if this exception is thrown? We should expect occasional transient download failures, they should probably retry once and then exit the process.

lib/src/sdk_manager.dart
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// For this class, the cwd should be the root of the project.
+class DownloadingSdk extends Sdk {
+  static const String kSdkPathName = 'dart-sdk';
+
+  String _version;
+
+  DownloadingSdk() {
+    _version = new File('dart-sdk.version').readAsStringSync().trim();
```
lukechurch: We probably need to update _version in common_server.dart to hand back this in response to version get requests rather than pulling it from the compiler string?

lib/src/sdk_manager.dart
```diff
@@ -0,0 +1,96 @@
+import 'dart:async';
```
lukechurch: nit: Missing copyright

bin/sdk.dart
```diff
@@ -0,0 +1,13 @@
+
```
lukechurch: nit: Missing copyright

bin/sdk.dart
```diff
@@ -0,0 +1,13 @@
...
+/// Update the sdk in `dart-sdk/` if necessary.
```
lukechurch: Could we call this file bin/update_sdk.dart? sdk.dart doesn't imply to me that it's going to download a new SDK

lib/src/sdk_manager.dart
devoncarew: done

bin/sdk.dart
devoncarew: sure, updated

bin/sdk.dart
devoncarew: done

lib/src/sdk_manager.dart
devoncarew: Make sense! done

lib/src/sdk_manager.dart
devoncarew: I believe gunzip works on `tar.gz` files (not zip files). Can we install unzip via apt-get or some such?

lib/src/sdk_manager.dart
devoncarew: done

lib/src/sdk_manager.dart
devoncarew: 👍 done

lib/src/sdk_manager.dart
devoncarew: Good idea; I added a separate method to do the download, that takes an optional (named) param for the retry count.

devoncarew:
Thanks for the review! PR updated to address all issues, modulo a question about `unzip`. Sounds like we'll either need to write around not having unzip on gae, or provision it on the instances (as part of a bootstrap script?). In any case, something to address one way or another.

devoncarew:
Holding off sgtm. I don't think this will be a big roadblock - there are probably a few ways to solve the issue (unzipping the sdks).

devoncarew:
@lukechurch, can you check that `apt-get install unzip` populates unzip on our managed vms? We should likely install this if not present as part of our bootup sequence.

lukechurch:
@devoncarew Thanks for this, now the GAE instances have unzip on them, I'll bring this in

devoncarew:
Great! Just to not, this PR rev's to `1.25.0-dev.11.0`; the latest released sdk is `1.25.0-dev.13.0`.

PR #321, opened 2017-08-22
Automatically restart analysis server

lukechurch:
This automatically restarts the analysis server if a request times out and improves logging

TBR @devoncarew

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -67,21 +67,24 @@ class CommonServer {
   AnalysisServerWrapper analysisServer;
   AnalysisServerWrapper analysisServerStrong;
 
-  CommonServer(String sdkPath, this.container, this.cache,
+  String sdkPath;
+
+  CommonServer(String this.sdkPath, this.container, this.cache,
       this.srcRequestRecorder, this.counter) {
     hierarchicalLoggingEnabled = true;
     _logger.level = Level.ALL;
+  }
 
+  Future init() async {
     pub = enablePackages ? new Pub() : new Pub.mock();
     compiler = new Compiler(sdkPath, pub);
     analysisServer = new AnalysisServerWrapper(sdkPath, strongMode: false);
     analysisServerStrong = new AnalysisServerWrapper(sdkPath);
-  }
 
-  Future init() async {
     await analysisServer.init();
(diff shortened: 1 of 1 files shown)


lib/src/common_server.dart
```diff
@@ -67,28 +67,32 @@ class CommonServer {
   AnalysisServerWrapper analysisServer;
   AnalysisServerWrapper analysisServerStrong;
 
-  CommonServer(String sdkPath, this.container, this.cache,
+  String sdkPath;
```
devoncarew: final?

lib/src/common_server.dart
```diff
@@ -67,28 +67,32 @@ class CommonServer {
...
+
+  CommonServer(String this.sdkPath, this.container, this.cache,
```
devoncarew: you can remove the `String` annotation

lib/src/common_server.dart
```diff
@@ -101,6 +105,17 @@ class CommonServer {
     await analysisServerStrong.warmup(useHtml: useHtml);
   }
 
+  Future restart() async {
+    _logger.warning('Restarting CommonServer');
+    await shutdown();
+    _logger.info('Analysis Servers shutdown');
```
devoncarew: sp: Server(s)

lib/src/common_server.dart
```diff
@@ -458,14 +475,20 @@ class CommonServer {
...
+          sources,
+          new Location()
+            ..sourceName = sourceName
+            ..offset = offset);
+      _logger.info(
+          'PERF: Computed completions in ${watch.elapsedMilliseconds}ms.');
+      return response;
+    } catch (e, st) {
+      _logger.severe('Error during _complete', e, st);
+      await restart();
```
devoncarew: 👍

lib/src/common_server.dart
lukechurch: thanks, will fix

lib/src/common_server.dart
lukechurch: thanks, will fix

lib/src/common_server.dart
lukechurch: Bizarrely there are actually two of them (strong mode and non-strong mode). I keep forgetting this as well.

PR #322, opened 2017-08-24
pass in the client id

devoncarew:
- pass the `--client-id` cli param into the analysis server

@lukechurch


lukechurch:
lgtm - merging

PR #323, opened 2017-08-24
Add unzip to the install packages

lukechurch:
@devoncarew TBR

It turned out to be pretty easy to add unzip


devoncarew:
👍 awesome!

PR #324, opened 2017-09-06
Pass args to analysis server

lukechurch:
Pass:
—dartpad argument

and wire through the sdkPath argument.

TBR @devoncarew @danrubel

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -62,7 +62,7 @@ class AnalysisServerWrapper {
         if (dumpServerMessages) _logger.info('--> $str');
       }
 
-      final serverArgs = [
+      List<String> serverArgs = <String>[
         '--dartpad',
         '--client-id=DartPad',
         '--client-version=${_sdkVersion}'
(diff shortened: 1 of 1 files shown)


lib/src/analysis_server.dart
```diff
@@ -62,11 +62,21 @@ class AnalysisServerWrapper {
         if (dumpServerMessages) _logger.info('--> $str');
       }
 
-      _init = AnalysisServer.create(
+      final serverArgs = [
+        '--dartpad',
```
devoncarew: Does the currently specified sdk support this flag? (I know it was added recently).

lib/src/analysis_server.dart
```diff
@@ -62,11 +62,21 @@ class AnalysisServerWrapper {
         if (dumpServerMessages) _logger.info('--> $str');
       }
 
-      _init = AnalysisServer.create(
+      final serverArgs = [
```
devoncarew: should likely specify the type arg here

lib/src/analysis_server.dart
lukechurch: It's supported as of dev-13, we're now serving against dev-16.

lib/src/analysis_server.dart
lukechurch: thanks, will do.

lukechurch:
Travis is now happy again

PR #326, opened 2017-09-06
Compiler from image

lukechurch:
@devoncarew @danrubel: Until we find the underlying cause of a problem need to use the compiler from the AppEngine image not from the downloaded SDK.

This makes that change, and also adds a test to the Analysis server code to ensure we're not offering completions on path imports

### DIFF
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -83,6 +83,7 @@ class Compiler {
 
       // Due to an issue with the VM we need to use the Dart2JS instance from
       // the image, not from the downloaded SDK
+      // TODO(#327): Use the compiler from the downloaded SDK
 
       final dartPath =  Platform.resolvedExecutable;
 
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -101,7 +101,6 @@ class Compiler {
       if (result.exitCode != 0) {
         CompilationResults results = new CompilationResults();
         results._problems.add(new CompilationProblem._(result.stdout));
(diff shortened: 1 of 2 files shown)


devoncarew:
Seems like a tough call; the current situation is that the dart2js snapshot has issues, and this PR would make the version of dart2js diverge from the other tools (analysis)? 1.24 vs 1.25-dev?

Is there an issue we could track for the dart2js appjit snapshot issue on app engine?

lib/src/compiler.dart
```diff
@@ -73,19 +73,35 @@ class Compiler {
...
-      File mainDart = new File(path.join(temp.path, kMainDart));
+      String compileTarget = path.join(temp.path, kMainDart);
+      File mainDart = new File(compileTarget);
       mainDart.writeAsStringSync(input);
 
       File mainJs = new File(path.join(temp.path, '${kMainDart}.js'));
       File mainSourceMap =
           new File(path.join(temp.path, '${kMainDart}.js.map'));
 
+      // Due to an issue with the VM we need to use the Dart2JS instance from
```
devoncarew: I'd track this change with a TODO and an issue.

lukechurch:
I agree it's sub-optimal. Let's put a tracking issue in place. I've got better at debugging these issues since this started, so it might be resolvable quickly.

lukechurch:
I can repro the compiler test failures locally. Will fix.

lib/src/compiler.dart
lukechurch: done

PR #332, opened 2017-12-08
Cleanup

kevmoo:


.travis.yml
```diff
@@ -1,5 +1,4 @@
 language: dart
 dart:
-  - dev
-sudo: false
+  - dev/release/1.25.0-dev.16.0
```
devoncarew: Why a specific release? We'll need to bump this frequently.

kevmoo: Lining up with https://github.com/dart-lang/dart-services/blob/master/dart-sdk.version

devoncarew: makes sense. can you add a comment to this file that when changing one version we should change the other? We'll want the same comment in the other location, but we'll probably need to extend the parser a bit to handle comments. Can you open an issue for that (I'll take a look at it at some point)?

Alternatively we could have a build process to validate the versions are kept in sync.

kevmoo:
PTAL

lib/src/sdk_manager.dart
```diff
@@ -52,11 +52,14 @@ class HostSdk extends Sdk {
...
+  final String _versionFull;
 
-  DownloadingSdk() {
-    _versionFull = new File('dart-sdk.version').readAsStringSync().trim();
-  }
+  DownloadingSdk()
+      : _versionFull = new File('dart-sdk.version')
+            .readAsLinesSync()
+            .map((line) => line.trim())
+            .where((line) => line.isNotEmpty && !line.startsWith('#'))
```
devoncarew: 👍 awesome

PR #333, opened 2018-01-31
Disable source recording

lukechurch:
@mit-mit @anders-sandholm @devoncarew

### DIFF
--- a/lib/services_gae.dart
+++ b/lib/services_gae.dart
@@ -56,7 +56,7 @@ class GaeServer {
     discoveryEnabled = false;
     fileRelayServer = new FileRelayServer();
     commonServer = new CommonServer(sdkPath, new GaeServerContainer(),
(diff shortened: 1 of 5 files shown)


lib/services_gae.dart
```diff
@@ -147,30 +147,6 @@ class GaeCache implements ServerCache {
...
-  @override
-  Future record(String verb, String source,
-      [int offset = -99, maxRetries = 3]) {
-    if (maxRetries < 0) {
-      _logger.warning("Soft-ERR: Max retries exceeded in SourceRequest.record");
-      return new Future.value(null);
-    }
-
-    int ms = new DateTime.now().millisecondsSinceEpoch;
-    GaeSourceRecordBlob record =
```
jakobr-google: Can we remove `GaeSourceRecordBlob`, too?

lukechurch: Thanks, yes, done.

PR #358, opened 2019-01-11
upgrade deps

devoncarew:
- upgrade several deps
- fix https://github.com/dart-lang/dart-services/issues/357

@jcollins-g @kevmoo

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -2,6 +2,7 @@ name: dart_services
 author: Dart Team <misc@dartlang.org>
 description: The server backend for a web based interactive Dart service.
 homepage: https://github.com/dart-lang/dart-services
+publish_to: none
 
 environment:
   sdk: '>=2.0.0-dev.64.1 <3.0.0'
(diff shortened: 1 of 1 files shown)


pubspec.yaml
```diff
@@ -7,19 +7,19 @@ environment:
   sdk: '>=2.0.0-dev.64.1 <3.0.0'
 
 dependencies:
-  _discoveryapis_commons: '>=0.1.0 <0.2.0'
```
kevmoo: While you're here: add a `publish_to: none` to this package – just to emphasize the point 😄

devoncarew: 👍 done

kevmoo: I doubled-down here, too

devoncarew: Hmm, I wouldn't necessarily have removed these myself...

devoncarew:
@jcollins-g, let me know if this is good to land, as this could impact dart-services behavior -

PR #373, opened 2019-03-05
Upgrade to Dart 2.2

domesticmouse:


devoncarew:
Is there a preview link we can try out? Or, one we can try about before the instance serving this version becomes the default?

devoncarew:
Ah, this version is live; nevermind.

PR #374, opened 2019-03-05
Excluding doc/generated/** from the analyzer

domesticmouse:


domesticmouse:
There is a step in the build process that generates the RPC bindings that are then copied into `dart-pad`. As it is generated code, we'd need to update the generator to make the code comply with the lint checks.

PR #375, opened 2019-03-07
Initial DDC support

vsmenon:


lib/src/compiler.dart
```diff
@@ -59,23 +60,23 @@ class Compiler {
...
       File mainDart = File(compileTarget);
+      // Hack for flutter.
+      input = input.replaceAll('package:flutter/', 'package:flutter_web/');
       mainDart.writeAsStringSync(input);
 
       File mainJs = File(path.join(temp.path, '${kMainDart}.js'));
       File mainSourceMap = File(path.join(temp.path, '${kMainDart}.js.map'));
 
-      final dart2JSPath = path.join(sdkPath, 'bin', 'dart2js');
+      final dart2JSPath = path.join(sdkPath, 'bin', 'dartdevc');
```
devoncarew: @vsmenon - so I understand - this only compiles the given dart source to a single js module, and returns that? A more complete solution would do this same compilation, but also return to the client information about the other modules to link in (require.js, the dart sdk, other large package modules, ...)?

vsmenon: Yes.

It's embedded in the source depending on the `modules` flag.   The `modules=amd` flag tells DDC to use AMD / require.js and encode dependences in the indicated module format.

I.e., it'll return something like:

```javascript
define(['dart_sdk', 'flutter'], function(dart_sdk, flutter) {
  ...
  main.main = function() {
    widgets.runApp(new src__widgets__basic.Center.new({child: new main.SpinningSquare.new()}));
  };
  ...
}
```

vsmenon: This is the full result of the hello sample:

```javascript
define(['dart_sdk'], function(dart_sdk) {
  'use strict';
  const core = dart_sdk.core;
  const dart = dart_sdk.dart;
  const dartx = dart_sdk.dartx;
  const _root = Object.create(null);
  const main = Object.create(_root);
  main.main = function() {
    for (let i = 0; i < 4; i++) {
      core.print("hello " + dart.str(i));
    }
  };
  dart.trackLibraries("main.dart", {
    "main.dart": main
  }, null);
  // Exports:
  return {
    main: main
  };
});

//# sourceMappingURL=main.dart.js.map
}
```

Note, other than `require.js`, those deps don't need to be added via script tags.  The require loader will find them and load them.

devoncarew: > The require loader will find them and load them.

How? The content isn't served from a server. I suspect that we'll need to have explicit script tags for the sdk dep, and any other compiled package artifacts.

vsmenon: Right.  Either:
- They need to be served along side the `frame.html` file as in this PR.  In that case, `require` will automatically find them.
- They need to be served in some well known path.  `require` can be configure to find it there: https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/tool/ddb#L163
- They need to be loaded as script tags.  

A potential advantage to having `require` do it, is that it would only load `flutter` if the sample used it.

devoncarew: Great, thanks for the info! Makes sense.

devoncarew:
We have enough of DDC support landed in this repo currently that I'll close this reference PR.

PR #377, opened 2019-03-12
Upgrading to Dart 2.2.1-dev.1.0

domesticmouse:


devoncarew:
This is for for deploying? Or Testing? We should probably keep DartPad on the latest stable, and not roll to a dev release.

If for trying out a dev release in between stable versions, perhaps we could support a `dev.dartpad.org` dev version?

domesticmouse:
A dartpad running developer version of dart sounds like a great idea. We have a variety of options on how to expose it, including URL parameters.

PR #378, opened 2019-03-20
add initial support for compiling with DDC

devoncarew:
- add a `compileDDC()` restful call
- remove an unsued `checked` param to the `compile()` request
- add tests for the compile DDC call
- regenerate the `doc/generated/dartservices.dart` / dartservices.json files
- refactor some of the request / response objects to improve the API a bit
- fixes https://github.com/dart-lang/dart-pad/issues/951

a follow up to https://github.com/dart-lang/dart-services/pull/375

(some unintentional diffs due to formatting existing files)

@RedBrogdon @domesticmouse

### DIFF
--- a/lib/src/api_classes.dart
+++ b/lib/src/api_classes.dart
@@ -90,12 +90,6 @@ class CompileRequest {
 
   @ApiProperty(
       description:
-          'Ignored. (server always assumes checked mode/asserts enabled)')
-  @deprecated
-  bool useCheckedMode;
-
-  @ApiProperty(
-      description:
           'Return the Dart to JS source map; optional (defaults to false).')
   bool returnSourceMap;
 }
--- a/lib/src/api_classes.dart
+++ b/lib/src/api_classes.dart
@@ -81,6 +81,7 @@ class Location {
   int offset;
 
   Location();
+
   Location.from(this.sourceName, this.offset);
 }
 
@@ -99,6 +100,18 @@ class CompileResponse {
   final String sourceMap;
 
   CompileResponse(this.result, [this.sourceMap]);
+}
+
+class CompileDDCRequest {
+  @ApiProperty(required: true, description: 'The Dart source.')
+  String source;
+}
+
+class CompileDDCResponse {
+  final String result;
+  final List<String> staticScriptUris;
+
+  CompileDDCResponse(this.result, this.staticScriptUris);
 }
 
 class CounterRequest {
@@ -165,6 +178,7 @@ class ProblemAndFixes {
   final int length;
 
   ProblemAndFixes() : this.fromList([]);
+
   ProblemAndFixes.fromList(
       [this.fixes, this.problemMessage, this.offset, this.length]);
 }
@@ -175,6 +189,7 @@ class CandidateFix {
   final List<SourceEdit> edits;
 
   CandidateFix() : this.fromEdits();
+
   CandidateFix.fromEdits([this.message, this.edits]);
 }
 
@@ -197,6 +212,7 @@ class SourceEdit {
   final String replacement;
 
   SourceEdit() : this.fromChanges();
+
   SourceEdit.fromChanges([this.offset, this.length, this.replacement]);
 
   String applyTo(String target) {
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -60,6 +60,7 @@ class RedisCache implements ServerCache {
 
   // Version of the server to add with keys.
   final String serverVersion;
+
   // pseudo-random is good enough.
   final Random randomSource = Random();
   static const int _connectionRetryBaseMs = 250;
@@ -71,17 +72,21 @@ class RedisCache implements ServerCache {
   }
 
   var _connected = Completer<void>();
+
   /// Completes when and if the redis server connects.  This future is reset
   /// on disconnection.  Mostly for testing.
   Future get connected => _connected.future;
 
   var _disconnected = Completer<void>()..complete();
+
   /// Completes when the server is disconnected (begins completed).  This
   /// future is reset on connection.  Mostly for testing.
   Future get disconnected => _disconnected.future;
 
   String __logPrefix;
-  String get _logPrefix => __logPrefix ??= 'RedisCache [${redisUriString}] (${serverVersion})';
+
+  String get _logPrefix =>
+      __logPrefix ??= 'RedisCache [${redisUriString}] (${serverVersion})';
 
   bool _isConnected() => redisClient != null && !_isShutdown;
   bool _isShutdown = false;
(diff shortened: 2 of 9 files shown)


devoncarew:
I'll need to follow this up with a PR that establishes a process for uploading the sdk js artifacts to google storage as part of a deploy, and that validates that an artifact exists in google storage for the current sdk version - checked as part of our travis job.

lib/src/common_server.dart
```diff
@@ -529,49 +557,101 @@ class CommonServer {
...
+    });
+  }
+
+  Future<CompileDDCResponse> _compileDDC(String source) async {
+    if (source == null) {
+      throw BadRequestError('Missing parameter: \'source\'');
+    }
+    String sourceHash = _hashSource(source);
+
+    // TODO(lukechurch): Remove this hack after
```
domesticmouse: Silly question, do we still need this hack? It looks like said issue has been resolved.

devoncarew: I don't think we do. From reading it, it looks like it was a way to disable caching for an older health service we had.

That service is no longer in use, and is we want to disable caching for some reason, we could add an optional param to the compile calls to do that.

devoncarew: I can clean that up in a follow-on PR.

lib/src/api_classes.dart
```diff
@@ -81,19 +81,14 @@ class Location {
...
   Location();
+
   Location.from(this.sourceName, this.offset);
 }
 
 class CompileRequest {
   @ApiProperty(required: true, description: 'The Dart source.')
   String source;
 
-  @ApiProperty(
```
RedBrogdon: Just curious: will this fail gracefully if an old front-end is still sending the property, or will dart-services and dart-pad need to be deployed together?

lib/src/common_server.dart
```diff
@@ -529,49 +557,101 @@ class CommonServer {
...
+    // https://github.com/dart-lang/rpc/issues/15 lands
+    String trimSrc = source.trim();
+    bool suppressCache = trimSrc.endsWith("/** Supress-Memcache **/") ||
+        trimSrc.endsWith("/** Suppress-Memcache **/");
+
+    String memCacheKey = "%%COMPILE_DDC:v0:source:$sourceHash";
```
RedBrogdon: We've talked about the need to update the static files (precompiled flutter_web, and Dart SDK) as part of each Flutter/dart release. Does any information relating to the versions of those files need to be included in the memcache key to avoid retrieving mismatched code while rolling out a new version of dart-services targeting new static files?

lib/src/common_server.dart
```diff
@@ -529,49 +557,101 @@ class CommonServer {
...
+    if (!suppressCache && result != null) {
+      log.info("CACHE: Cache hit for compile");
+      var resultObj = JsonDecoder().convert(result);
+      return CompileResponse(
+        resultObj["compiledJS"],
+        returnSourceMap ? resultObj["sourceMap"] : null,
+      );
+    }
+
+    log.info("CACHE: MISS, forced: $suppressCache");
```
RedBrogdon: Nit: The log statements for cache hits reference which method (compile/compileDDC) was being used, but not the ones for cache misses. Should those be updated?

lib/src/common_server.dart
```diff
@@ -529,49 +557,101 @@ class CommonServer {
...
+        unawaited(setCache(memCacheKey, cachedResult));
+
+        return CompileDDCResponse(results.compiledJS, results.staticScriptUris);
+      } else {
+        List<CompilationProblem> problems = results.problems;
+        String errors = problems.map(_printCompileProblem).join('\n');
+        throw BadRequestError(errors);
       }
+    }).catchError((e, st) {
+      log.severe('Error during compile: ${e}\n${st}');
```
RedBrogdon: nit: "...during compileDDC?"

lib/src/common_server.dart
```diff
@@ -529,49 +557,101 @@ class CommonServer {
...
+
+    log.info("CACHE: MISS, forced: $suppressCache");
+    Stopwatch watch = Stopwatch()..start();
+
+    return compiler.compileDDC(source).then((DDCCompilationResults results) {
+      if (results.hasOutput) {
+        int lineCount = source.split('\n').length;
+        int outputSize = (results.compiledJS.length + 512) ~/ 1024;
+        int ms = watch.elapsedMilliseconds;
+        log.info('PERF: Compiled ${lineCount} lines of Dart into '
```
RedBrogdon: nit: Reference CompileDDC here?

lib/src/compiler.dart
```diff
@@ -75,15 +78,73 @@ class Compiler {
...
+      arguments.addAll(['-o', '${kMainDart}.js']);
+      arguments.add(kMainDart);
+
+      String compileTarget = path.join(temp.path, kMainDart);
+      File mainDart = File(compileTarget);
+      mainDart.writeAsStringSync(input);
+
+      File mainJs = File(path.join(temp.path, '${kMainDart}.js'));
+
+      final dart2JSPath = path.join(sdkPath, 'bin', 'dartdevc');
```
RedBrogdon: nit: rename dart2JS references here?

test/common_server_test.dart
```diff
@@ -84,12 +84,13 @@ void defineTests() {
     // since they talk to the same redis instances.
     Lock singleTestOnly = Lock();
 
-    // Prevent cases where we might try to reenter addStream for either
-    // stdout or stderr (which will throw a BadState).
+    // Prevent cases where we might try to reenter addStream for either  stdout
```
RedBrogdon: super-nit: double space in front of stdout

RedBrogdon:
LGTM. All the comments are all either nits or questions about things I'd like to make sure I understand. 😄

lib/src/api_classes.dart
devoncarew: Yes - they'll come is as additional restful params (as a POST or a GET), and just be ignored - they won't be passed along to the lower level dart apis.

lib/src/common_server.dart
devoncarew: I'll do that for the log statement in changed or new code. We may want to re-think our logging, to make sure that we're logging things we want to track (this hasn't been updated or re-thought in a while).

lib/src/common_server.dart
devoncarew: Good question. From looking at the code, our redis impl already factors in the sdk version for all cache items. I added a todo in this code that we'll need to include the version of major packages used in the cache keys, once we start supporting packages.

lib/src/common_server.dart
devoncarew: done

lib/src/common_server.dart
devoncarew: done

lib/src/compiler.dart
devoncarew: oops

test/common_server_test.dart
devoncarew: done

PR #381, opened 2019-03-27
Up memory allocation & fix compilation 500 errors

domesticmouse:


domesticmouse:
PTAL @RedBrogdon =)

pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   pedantic: ^1.4.0
   shelf: ^0.7.0
   rpc: ^0.6.1
-  uuid: 2.0.0-rc1
+  uuid: 2.0.0
```
RedBrogdon: Does this need to be pinned to generate stable uuids?

devoncarew: I believe the uuid generated are probably the same with either version, and the algorithm we're using this for really just needs the same keys for the same input per session.

PR #382, opened 2019-03-29
Aligning lint rules with dart-pad

domesticmouse:


lib/src/compiler.dart
```diff
@@ -95,7 +95,7 @@ class Compiler {
         return results;
       }
     } catch (e, st) {
-      _logger.warning("Compiler failed: $e /n $st");
```
domesticmouse: I doubt this was intended

lib/src/compiler.dart
```diff
@@ -149,7 +149,7 @@ class Compiler {
...
```
domesticmouse: Nor this

tool/grind.dart
```diff
@@ -94,19 +94,19 @@ void discovery() {
   File discoveryFile = File('doc/generated/dartservices.json');
   discoveryFile.parent.createSync();
   log('writing ${discoveryFile.path}');
-  discoveryFile.writeAsStringSync(result.stdout.trim() + '\n');
+  discoveryFile.writeAsStringSync('${result.stdout.trim()}\n');
 
   ProcessResult resultDb = Process.runSync(
       Platform.executable, ['bin/server_dev.dart', '--discovery', '--relay']);
 
-  if (result.exitCode != 0) {
```
domesticmouse: This was constant false, due to prior throw. I'm guessing cut'n'paste fail.

RedBrogdon:
This LGTM, but it might be a good one to hold off for a couple weeks until Devon is done with his changes, just to avoid slowing him down.

devoncarew:
I can work around this, so no problems w/ landing it while I'm still working on my PR.

RedBrogdon:
SGTM, then!

PR #383, opened 2019-04-01
some fixes to the DDC support

devoncarew:
- create well-formed AMD modules (they need to be named for us to be able to load them in the front end)
- change the compileDDC endpoint API - instead of the list of all the artifacts, we just need to know the directory containing them. In the future, the directory would be different for the Dart sdk version we're compiling for

@RedBrogdon @domesticmouse

### DIFF
--- a/doc/generated/dartservices.dart
+++ b/doc/generated/dartservices.dart
@@ -814,29 +814,28 @@ class CandidateFix {
 }
 
 class CompileDDCResponse {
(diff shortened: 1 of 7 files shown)


devoncarew:
@jakemac53 for fyi

devoncarew:
Updated to use the `--single-out-file` arg to DDC (instead of a regex to create a module with a name).

PR #384, opened 2019-04-10
build the necessary fluter_web artifacts

devoncarew:
This PR:
- contributes a grind task to build flutter_web artifacts
- adds validation to our CI to ensure that the flutter_web artifacts exist for the currently supported version of the dart sdk

The code has been modified slightly in order to split a very large PR up into smaller ones for review.

@RedBrogdon @domesticmouse

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -91,7 +91,7 @@ void validateStorageArtifacts() async {
 }
 
 Future _validateExists(String url) async {
-  log(url);
+  log('checking $url...');
 
   http.Response response = await http.head(url);
   if (response.statusCode != 200) {
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -3,13 +3,16 @@
 // BSD-style license that can be found in the LICENSE file.
 
 // ignore_for_file: always_specify_types
-
 library services.grind;
 
 import 'dart:async';
 import 'dart:io';
 
+// TODO(devoncarew): re-add this import after review
+//import 'package:dart_services/src/flutter_web.dart';
 import 'package:grinder/grinder.dart';
+import 'package:grinder/grinder_files.dart';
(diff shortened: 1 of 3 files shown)


tool/grind.dart
```diff
@@ -61,11 +65,130 @@ void updateDockerVersion() {
...
+  joinFile(dir, ['pubspec.yaml']).writeAsStringSync(pubspec);
+
+  // run pub get
+  Pub.get(workingDirectory: dir.path);
+
+  // locate the artifacts
+  final List<String> flutterPackages = [
+    'flutter_web',
+    'flutter_web_ui',
+    // TODO(devoncarew): package:flutter_web_test is not currently handled
```
RedBrogdon: Is this coming in a a subsequent PR? I'm going to need it to run widget tests.

tool/grind.dart
```diff
@@ -61,11 +65,130 @@ void updateDockerVersion() {
...
+  const String urlBase =
+      'https://storage.googleapis.com/compilation_artifacts/';
+
+  for (String artifact in compilationArtifacts) {
+    await _validateExists('$urlBase$version/$artifact');
+  }
+}
+
+Future _validateExists(String url) async {
+  log(url);
```
RedBrogdon: nit: consider a more meaningful message like 'Checking to see if $url exists', or just move this into an else block in the if-then below and report success.

tool/grind.dart
devoncarew: Yup, we'll need to support this, I just ran into issues and am trying to limit how many I'm solving at one time.

RedBrogdon: SGTM.

tool/grind.dart
devoncarew: 👍 will do

PR #385, opened 2019-04-10
add support for compiling and analyzing package:flutter_web apps

devoncarew:
add support for compiling and analyzing package:flutter_web apps:
- simplify the existing abstraction around package imports
- add a new abstraction in to support `package:flutter_web/` and related imports
- provision the package once per server, the first time we get a flutter_web request in
- fix https://github.com/dart-lang/dart-pad/issues/954, fix https://github.com/dart-lang/dart-pad/issues/953
- add tests
- add a local config override, so you can point to a local copy of package:flutter_web

@domesticmouse @RedBrogdon 

I may need to disable some of the tests - they likely depend on being able to provision package:flutter_web.


devoncarew:
(updated tests and the grind script)

lib/src/common_server.dart
```diff
@@ -545,6 +547,8 @@ class CommonServer {
       return results;
     } catch (e, st) {
       log.severe('Error during analyze', e, st);
+      // TODO(devoncarew): It's really not clear to me that we should restart
```
domesticmouse: Do we risk leaking information between runs if we don't restart?

devoncarew: I think that's a good point - we can know what state the analysis server is in. I'll remove the comment.

lib/src/flutter_web.dart
```diff
@@ -0,0 +1,202 @@
...
+        (String prefix) => import.startsWith(prefix),
+      );
+    });
+  }
+
+  bool hasUnsupportedImport(Set<String> imports) {
+    return getUnsupportedImport(imports) != null;
+  }
+
+  String getUnsupportedImport(Set<String> imports) {
```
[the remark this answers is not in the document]
devoncarew: I'm just trying to provide a reasonable failure message - in this case we show the first invalid import. There are likely UI improvements to be made here - like underlying the imports instead.

In this case I don't want to pass the imports through to the lower levels (analysis, compilation), so we won't get better error feedback w/o writing some code.

If you think we should report all invalid imports, we can track that in a todo or an issue.

devoncarew:
Travis failures are due to the latest dart dev sdk; I filed https://github.com/dart-lang/dart-pad/issues/1002 to track.

I'll remove testing on dev in this PR to get it green.

PR #386, opened 2019-04-11
Remove lint always_specify_types

natebosch:
This lint is not appropriate for repos under `dart-lang`. Only the
flutter SDK should use this lint.

For repositories under `dart-lang` we should always follow
https://www.dartlang.org/guides/language/effective-dart

Also run `dartfmt`.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -9,7 +9,6 @@ analyzer:
 linter:
   rules:
     - always_declare_return_types
-    - always_specify_types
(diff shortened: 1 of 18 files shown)


natebosch:
cc @RedBrogdon @domesticmouse - note that `dart-lang` does _not_ follow the same guidelines as `flutter`

RedBrogdon:
LGTM. I'll definitely take your word on Dart best practices.

PR #387, opened 2019-04-11
update to use dartdevc as a persistent worker

jakemac53:
Closes https://github.com/dart-lang/dart-pad/issues/1003

cc @devoncarew @RedBrogdon @domesticmouse not sure what combination of you would like to review this :).

This uses the bazel_worker package to run a pool of persistent dartdevc tasks and schedule jobs between them. Currently it is using the default number of workers (4), but that is easily tweakable (and I don't know what makes the most sense for dartpad - if it only handles one request at a time then just 1 would be needed although it doesn't hurt to have it higher technically - it would only spin up 1 in that case still). 

Note that I did not switch to using kernel - that would be a bit more involved and doesn't really seem necessary although we could get additional wins from it.

I did a few other cleanup things as well, feel free to push back on any of those, I will leave specific comments as to why I changed them though.

I added general dev compiler benchmarks and a spinning square benchmark for all modes and got the following results for ddc:

## Before
```
[dartdevc.hello            :  181.118ms]
[dartdevc.hellohtml        :  219.658ms]
[dartdevc.sunflower        :  241.977ms]
[dartdevc.spinning_square  :  251.619ms]
```

## After

```
[dartdevc.hello            :   19.113ms]
[dartdevc.hellohtml        :   32.594ms]
[dartdevc.sunflower        :   46.399ms]
[dartdevc.spinning_square  :   41.897ms]
```

### DIFF
--- a/benchmark/bench.dart
+++ b/benchmark/bench.dart
@@ -8,32 +8,51 @@ library services.bench;
 
 import 'dart:async';
 
+import 'package:logging/logging.dart';
+
 import 'package:dart_services/src/analysis_server.dart';
 import 'package:dart_services/src/bench.dart';
 import 'package:dart_services/src/common.dart';
 import 'package:dart_services/src/compiler.dart';
 import 'package:dart_services/src/flutter_web.dart';
 
-void main(List<String> args) {
+void main(List<String> args) async {
   final bool json = args.contains('--json');
 
   final BenchmarkHarness harness = BenchmarkHarness(asJson: json);
 
   final FlutterWebManager flutterWebManager = FlutterWebManager(sdkPath);
+  await flutterWebManager.warmup();
+  var compiler = Compiler(sdkPath, flutterWebManager);
+
+  Logger.root.level = Level.WARNING;
+  Logger.root.onRecord.listen((LogRecord record) {
+    print(record);
+    if (record.stackTrace != null) print(record.stackTrace);
+  });
 
   final List<Benchmark> benchmarks = [
     AnalyzerBenchmark('hello', sampleCode, flutterWebManager),
     AnalyzerBenchmark('hellohtml', sampleCodeWeb, flutterWebManager),
     AnalyzerBenchmark('sunflower', _sunflower, flutterWebManager),
+    AnalyzerBenchmark('spinning_square', _spinningSquare, flutterWebManager),
     AnalysisServerBenchmark('hello', sampleCode, flutterWebManager),
     AnalysisServerBenchmark('hellohtml', sampleCodeWeb, flutterWebManager),
     AnalysisServerBenchmark('sunflower', _sunflower, flutterWebManager),
-    Dart2jsBenchmark('hello', sampleCode, flutterWebManager),
-    Dart2jsBenchmark('hellohtml', sampleCodeWeb, flutterWebManager),
-    Dart2jsBenchmark('sunflower', _sunflower, flutterWebManager),
+    AnalysisServerBenchmark(
+        'spinning_square', _spinningSquare, flutterWebManager),
+    Dart2jsBenchmark('hello', sampleCode, compiler),
+    Dart2jsBenchmark('hellohtml', sampleCodeWeb, compiler),
+    Dart2jsBenchmark('sunflower', _sunflower, compiler),
+    Dart2jsBenchmark('spinning_square', _spinningSquare, compiler),
+    DevCompilerBenchmark('hello', sampleCode, compiler),
+    DevCompilerBenchmark('hellohtml', sampleCodeWeb, compiler),
+    DevCompilerBenchmark('sunflower', _sunflower, compiler),
+    DevCompilerBenchmark('spinning_square', _spinningSquare, compiler),
   ];
 
-  harness.benchmark(benchmarks);
+  await harness.benchmark(benchmarks);
+  await compiler.dispose();
 }
 
 class AnalyzerBenchmark extends Benchmark {
@@ -60,14 +79,27 @@ class Dart2jsBenchmark extends Benchmark {
   final String source;
   final Compiler compiler;
 
-  Dart2jsBenchmark(
-      String name, this.source, FlutterWebManager flutterWebManager)
-      : compiler = Compiler(sdkPath, flutterWebManager),
-        super('dart2js.$name');
+  Dart2jsBenchmark(String name, this.source, this.compiler)
+      : super('dart2js.$name');
 
   @override
   Future perform() {
     return compiler.compile(source).then((CompilationResults result) {
+      if (!result.success) throw result;
+    });
+  }
+}
+
+class DevCompilerBenchmark extends Benchmark {
+  final String source;
+  final Compiler compiler;
+
+  DevCompilerBenchmark(String name, this.source, this.compiler)
+      : super('dartdevc.$name');
+
+  @override
+  Future perform() {
+    return compiler.compileDDC(source).then((DDCCompilationResults result) {
       if (!result.success) throw result;
     });
   }
@@ -106,7 +138,7 @@ class Sunflower {
   static const String ORANGE = "orange";
   static const SEED_RADIUS = 2;
   static const SCALE_FACTOR = 4;
-  static const TAU = math.PI * 2;
+  static const TAU = math.pi * 2;
   static const MAX_D = 300;
 
   CanvasRenderingContext2D ctx;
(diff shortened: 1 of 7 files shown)


benchmark/bench.dart
```diff
@@ -8,32 +8,51 @@ library services.bench;
...
   final bool json = args.contains('--json');
 
   final BenchmarkHarness harness = BenchmarkHarness(asJson: json);
 
   final FlutterWebManager flutterWebManager = FlutterWebManager(sdkPath);
+  await flutterWebManager.warmup();
+  var compiler = Compiler(sdkPath, flutterWebManager);
+
+  Logger.root.level = Level.WARNING;
+  Logger.root.onRecord.listen((LogRecord record) {
```
jakemac53: This just helps with diagnosing issues when running the benchmarks (I didn't realize I had to set up the config file for flutter_web, and just got a failure with no output before)

benchmark/bench.dart
```diff
@@ -8,32 +8,51 @@ library services.bench;
...
 import 'package:dart_services/src/flutter_web.dart';
 
-void main(List<String> args) {
+void main(List<String> args) async {
   final bool json = args.contains('--json');
 
   final BenchmarkHarness harness = BenchmarkHarness(asJson: json);
 
   final FlutterWebManager flutterWebManager = FlutterWebManager(sdkPath);
+  await flutterWebManager.warmup();
```
jakemac53: This fetches the flutter web summary files etc

lib/src/compiler.dart
```diff
@@ -185,6 +190,11 @@ class CompilationResults {
 
   /// This is true if there were no errors.
   bool get success => problems.isEmpty;
+
+  @override
+  String toString() => success
```
jakemac53: This also helps with debugging compilation errors (previous toString was just the runtime type of the exception)

lib/src/flutter_web.dart
```diff
@@ -52,11 +52,7 @@ $_samplePackageName:lib/
   }
 
   Future<void> warmup() async {
-    try {
-      await initFlutterWeb();
-    } catch (_) {
-      // ignore any exception here
-    }
+    await initFlutterWeb();
```
jakemac53: This try/catch was hiding the flutter setup issues also so I removed it - but can add back in if we need it as a safety check or something? I think it should at least log a warning or severe message though.

benchmark/bench.dart
domesticmouse: I feel like this should be a code comment somewhere such that this knowledge doesn't disappear on submit.

lib/src/flutter_web.dart
domesticmouse: I really dislike catch and ignore. Happy to try living without it for the time being.

benchmark/bench.dart
jakemac53: done

lib/src/flutter_web.dart
devoncarew: We shouldn't propagate any exception from here - warmup is meant to be a best effort, and not stop the initialization of the server. If this call to `initFlutterWeb` fails, a future one (reacting to a user request) could succeed.

I'd catch any exception and log it. Definitely agree that silently catching and ignoring is a good place for bugs to hide.

benchmark/bench.dart
```diff
@@ -8,32 +8,52 @@ library services.bench;
...
 
-void main(List<String> args) {
+void main(List<String> args) async {
   final bool json = args.contains('--json');
 
   final BenchmarkHarness harness = BenchmarkHarness(asJson: json);
 
   final FlutterWebManager flutterWebManager = FlutterWebManager(sdkPath);
+  // Fetches the flutter web summary files etc from google cloud storage.
+  await flutterWebManager.warmup();
```
devoncarew: If the intent here is to make the flutter web manager class ready to benchmark flutter compiles, you may want to instead call `initFlutterWeb` directly.

devoncarew: Which `warmup` does today, but `initFlutterWeb` is a bit more semantic.

lib/src/flutter_web.dart
jakemac53: done, logging a warning now

benchmark/bench.dart
jakemac53: Done, I also removed the comment since it is more self explanitory now.

jakemac53:
> This looks great to me. The only thing I want to check is whether the existing mechanism that kept dart-services from trying to start two compiles at once targeting the same location will also work here. Is there anything changing that could cause state from one compileDDC request to bleed over into another or cause two processes kicked off by _ddcDriver to fight over a shared root directory for the compile?

This does work slightly differently, but it has the same behavior. The difference is that since the [processes](https://github.com/dart-lang/dart-services/pull/387/files#diff-6f8db5c7de455147a9ca9a2f78b305c2R32) are persistent they all run with the same current working directory (current just the cwd of the main process, but we could create a different temp dir if you would like and run them there for additional safety).

Each compile call still creates its [own temp directory](https://github.com/dart-lang/dart-services/pull/387/files#diff-6f8db5c7de455147a9ca9a2f78b305c2R128), where it writes the [main.dart](https://github.com/dart-lang/dart-services/pull/387/files#diff-6f8db5c7de455147a9ca9a2f78b305c2R139), and we pass the [full path](https://github.com/dart-lang/dart-services/pull/387/files#diff-6f8db5c7de455147a9ca9a2f78b305c2R146) to that to dartdevc, as well as making the [output path](https://github.com/dart-lang/dart-services/pull/387/files#diff-6f8db5c7de455147a9ca9a2f78b305c2R143) be under that directory as well.

domesticmouse:
Thanks all!

devoncarew:
I see this issue now when using the `compileDDC` endpoint locally:

```
VM2060:52 Uncaught TypeError: Cannot read property 'main' of undefined
    at <anonymous>:52:23
    at Object.execCb (VM1379 require.js:1696)
    at Module.check (VM1379 require.js:883)
    at Module.<anonymous> (VM1379 require.js:1139)
    at VM1379 require.js:134
    at VM1379 require.js:1189
    at each (VM1379 require.js:59)
    at Module.emit (VM1379 require.js:1188)
    at Module.check (VM1379 require.js:938)
    at Module.enable (VM1379 require.js:1176)
    at Module.init (VM1379 require.js:788)
    at callGetModule (VM1379 require.js:1203)
    at Object.completeLoad (VM1379 require.js:1590)
    at HTMLScriptElement.onScriptLoad (VM1379 require.js:1717)
```

I haven't investigated to narrow down the issue.

devoncarew:
(I haven't bisected either, to narrow it down to this PR, but suspect it is this one)

domesticmouse:
I'm guessing we should add tests on the `compileDDC` end point right about now?

devoncarew:
We have some to validate that the end point produces output, but not integration tests that the output can execute and produce expected results...

domesticmouse:
I just confirmed this is the change that caused the issues seen in https://github.com/dart-lang/dart-services/pull/387#issuecomment-483061815

I'm about to file a revert for this PR.

jakemac53:
Hmm my guess is that the module name changed or something?

jakemac53:
I would be happy to look at it if somebody can point me at some docs for how to set up the full stack for development :).

jakemac53:
Actually it looks like https://github.com/dart-lang/dart-pad/blob/master/CONTRIBUTING.md has some instructions, ill see if I can work out the issue here.

jakemac53:
hmm I can't get that fully to work (doesn't seem to actually start the server)

devoncarew:
I'm generally following those instructions. From the dart-pad directory, I run:

```
grind serve-custom-backend
```

and from the dart-services directory I run:

```
dart bin/server_dev.dart --port 8002
```

jakemac53:
Ah using `dart bin/server_dev.dart --port 8002` works (instead of `grind serve` which doesn't seem to work for me).

I noticed it still seems to be using dart2js though - how do i switch it to use the dev compiler?

devoncarew:
If testing with the main dartpad UI (the desktop version), you'll need to toggle the compile time flag in lib/playground.dart - `_useDDC`.

jakemac53:
Ok - still getting some errors fetching the `dart_sdk.js` file (loading the `dart_sdk` module) but I was able to at least see the compiled artifact and this is almost certainly the problem:

```
// Exports:
return {
  ________________________tmp__dartpadESMJSE__main: ________________________tmp__dartpadESMJSE__main
};
```

Dartpad is just expecting that to look like this:

// Exports:
return {
  main: main
};
```

I think I know how to fix that (pass --module-root).

PR #388, opened 2019-04-11
Eliminating implicit casts

domesticmouse:
At the suggestion of @natebosch, this PR restricts implicit casts to make sure we don't get back into the case where we error out with List<String> not being a sub-type of List<Object>.

Thoughts? 

/cc @RedBrogdon

### DIFF
--- a/lib/services_dev.dart
+++ b/lib/services_dev.dart
@@ -31,7 +31,7 @@ void main(List<String> args) {
   parser.addOption('server-url', defaultsTo: 'http://localhost');
 
   ArgResults result = parser.parse(args);
-  num port = int.tryParse(result['port']);
+  num port = int.tryParse(result['port'] as String);
   if (port == null) {
     stdout.writeln(
         'Could not parse port value "${result['port']}" into a number.');
@@ -45,9 +45,9 @@ void main(List<String> args) {
     exit(0);
   }
 
-  if (result['discovery']) {
-    String serverUrl = result['server-url'];
-    if (result['relay']) {
+  if (result['discovery'] as bool) {
+    String serverUrl = result['server-url'] as String;
+    if (result['relay'] as bool) {
       EndpointsServer.generateRelayDiscovery(sdk, serverUrl).then(printExit);
     } else {
       EndpointsServer.generateDiscovery(sdk, serverUrl).then(printExit);
@@ -61,7 +61,7 @@ void main(List<String> args) {
     if (record.stackTrace != null) print(record.stackTrace);
   });
 
-  EndpointsServer.serve(sdk, port).then((EndpointsServer server) {
+  EndpointsServer.serve(sdk, port.toInt()).then((EndpointsServer server) {
     _logger.info('Listening on port ${server.port}');
   });
 }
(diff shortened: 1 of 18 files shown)


lib/src/api_classes.dart
```diff
@@ -153,13 +153,13 @@ class CompleteResponse {
   static List<Map<String, String>> _convert(List<Map<dynamic, dynamic>> list) {
     return list.map<Map<String, String>>((Map<dynamic, dynamic> m) {
       Map<String, String> newMap = <String, String>{};
-      for (String key in m.keys) {
+      for (dynamic key in m.keys) {
```
natebosch: [optional] This could be

`for (var key in m.keys.cast<String>())`

that will be an explicit cast and still give the `key` variable the type `String` to make it so you don't need the change below.

lib/src/bench.dart
```diff
@@ -132,6 +132,7 @@ bool isCheckedMode() {
   try {
     String result = 'foo';
     result = '${result}bar';
+    // ignore: invalid_assignment
```
natebosch: This feels weird. There is also no such thing as unchecked mode anymore so we probably want to remove this method altogether. Maybe file a followup issue?

domesticmouse: I'll happily remove it, either in this PR or in a follow-up PR. I'm going to be living in this code base in the lead up to I/O, so the cleaner I can make it the better 😁

test/gae_deployed_test.dart
```diff
@@ -24,7 +24,7 @@ void defineTests({bool skip = true}) {
 
 void analyzeTest() {
   final String url = '$serverUrl/api/analyze';
-  Map headers = {'Content-Type': 'text/plain; charset=UTF-8'};
+  Map<String,String> headers = {'Content-Type': 'text/plain; charset=UTF-8'};
```
RedBrogdon: Would this also solve the cast problem?

```
final headers = <String, String>{ ... };
```

That's the construction I'm more used to seeing.

natebosch: Don't even need to types on the map once we remove the lint.
`final headers = {'Content-Type': 'text/plain; charset=UTF-8'};` would be most idiomatic.

lib/src/api_classes.dart
domesticmouse: done

lib/src/bench.dart
domesticmouse: Removed isCheckedMode check

lib/services_dev.dart
```diff
@@ -31,7 +31,7 @@ void main(List<String> args) {
   parser.addOption('server-url', defaultsTo: 'http://localhost');
 
   ArgResults result = parser.parse(args);
-  num port = int.tryParse(result['port']);
+  num port = int.tryParse(result['port'] as String);
```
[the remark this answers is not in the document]
domesticmouse: done

test/gae_deployed_test.dart
domesticmouse: done

PR #391, opened 2019-04-15
Revert "update to use dartdevc as a persistent worker (#387)"

domesticmouse:
This reverts commit 7371bbabcb86998b0ba80d2a7da0e7e0d0124a85.

### DIFF
--- a/benchmark/bench.dart
+++ b/benchmark/bench.dart
@@ -6,52 +6,32 @@ library services.bench;
 
 import 'dart:async';
 
-import 'package:logging/logging.dart';
-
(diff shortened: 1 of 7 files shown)


domesticmouse:
This revert is to deal with the issues reported in https://github.com/dart-lang/dart-services/pull/387#issuecomment-483061815

jakemac53:
See  https://github.com/dart-lang/dart-services/pull/392 which fixes the underlying issue as well

jakemac53:
Ok that is submitted now - I think you should be able to close this issue (I was able to compile and run apps with a local setup)

PR #395, opened 2019-04-16
Adding flutter_web to the Docker image

domesticmouse:


lib/src/common_server.dart
```diff
@@ -618,22 +618,22 @@ class CommonServer {
 
     await _checkPackageReferencesInitFlutterWeb(source);
 
-    String sourceHash = _hashSource(source);
+    // String sourceHash = _hashSource(source);
```
devoncarew: The guidance for the flutter/flutter codebase is to not commit checked out code. I've more or less gotten used to that - it's hard for the commented out code to bit rot for instance if its been deleted.

domesticmouse: Commented out code deleted

PR #397, opened 2019-04-18
Excluding flutter_web from analysis

domesticmouse:


RedBrogdon:
What's the reason behind the change? Not that I think there isn't one, just that I don't know enough to know what it is. :)

domesticmouse:
Dropping `analyze` because the presence of `flutter_web` breaks it, and `PORT` was part of my attempt to get `dart-services` running on [Cloud Run](https://cloud.google.com/run/). If someone can figure out why it is `exit(0)`ing, then we get auto-scaling.

PR #400, opened 2019-04-24
use unauthenticated cloud storage for dart sdk

johnpryan:
fixes dart-lang/dart-pad#1018


devoncarew:
Here's the url which the backend returns:

```
https://storage.googleapis.google.com/compilation_artifacts/2.2.0/dart_sdk.js
```

That gives a 404; perhaps we need a different base url?

johnpryan:
oops, still had working changes. Yes, the `.google` isn't supposed to be there

RedBrogdon:
Nice work!

PR #402, opened 2019-04-30
Probe to see if #1002 is still an issue

domesticmouse:
- fix https://github.com/dart-lang/dart-pad/issues/1002


domesticmouse:
PTAL @devoncarew, looks like https://github.com/dart-lang/dart-pad/issues/1002 is fixed.

devoncarew:
Great! Looks good then.

PR #403, opened 2019-04-30
filter out completion results that don't start with the replacement text

johnpryan:
I'm not sure if this is going to work for all scenarios... I've done some basic testing with the new embeddable editor

Screen Shot 2019-04-30 at 3 24 47 PM

closes dart-lang/dart-pad/#1014

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -332,7 +332,10 @@ class AnalysisServerWrapper {
       );
       CompletionResults results = await getCompletionResults(id.id);
       await _unloadSources();
-      return results;
+      var source = sources[_getPathFromName(sourceName)];
+      var prefix = source.substring(results.replacementOffset,
+          results.replacementOffset + results.replacementLength);
+      return _completionResultsOnlyWithPrefix(results, prefix);
     }, timeoutDuration: _ANALYSIS_SERVER_TIMEOUT));
   }
 
@@ -439,6 +442,15 @@ class AnalysisServerWrapper {
         }
       }
     });
+  }
+
+  CompletionResults _completionResultsOnlyWithPrefix(
+      CompletionResults results, String prevPattern) {
+    var filteredResults = results.results
+        .where((s) => s.completion.startsWith(prevPattern))
+        .toList();
+    return CompletionResults(results.id, results.replacementOffset,
+        results.replacementLength, filteredResults, results.isLast);
   }
 
   Future<CompletionResults> getCompletionResults(String id) {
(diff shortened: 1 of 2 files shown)


test/analysis_server_test.dart
```diff
@@ -176,6 +183,20 @@ void defineTests() {
       AnalysisResults results = await analysisServer.analyze(sampleDart2OK);
       expect(results.issues, hasLength(0));
     });
+
+    test('filter completions', () async {
+      // just after A
+      var idx = 61;
+      expect(completionLargeNamespaces.substring(idx - 1, idx), 'A');
+      return analysisServer
```
devoncarew: use `await` here instead of `then`?

lib/src/analysis_server.dart
```diff
@@ -332,7 +332,10 @@ class AnalysisServerWrapper {
       );
       CompletionResults results = await getCompletionResults(id.id);
       await _unloadSources();
-      return results;
+      var source = sources[_getPathFromName(sourceName)];
```
devoncarew: Perhaps move the filtering to `completeMulti()`? You then wouldn't have to create a new copy of `CompletionResults`, which is really an impl class from the analysis server library.

lib/src/analysis_server.dart
```diff
@@ -332,7 +332,10 @@ class AnalysisServerWrapper {
...
+      var prefix = source.substring(results.replacementOffset,
```
devoncarew: For `prefix`, we may want the string from `results.replacementOffset` to `offset`. `replacementLength` can extend past the cursor position I believe.

https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html, under 'completion.results'.

lib/src/analysis_server.dart
```diff
@@ -125,10 +125,13 @@ class AnalysisServerWrapper {
         await _completeImpl(sources, location.sourceName, location.offset);
     List<CompletionSuggestion> suggestions = results.results;
 
-    // This hack filters out of scope completions. It needs removing when we
-    // have categories of completions.
-    // TODO(devoncarew): Remove this filter code.
+    var source = sources[location.sourceName];
+    var prefix = source.substring(results.replacementOffset, location.offset);
     suggestions = suggestions
+        .where((s) => s.completion.startsWith(prefix))
```
devoncarew: we may want to compare case insensitive here (via toLowerCase on each completion, and once on `prefix`)

johnpryan:
added the case-insensitive piece. it's certainly fast on when the server is running locally :)

PR #410, opened 2019-05-22
Vendoring in flutter_web

domesticmouse:
Using the power of Git submodules, we now have a non hacky way to include `flutter_web` into the build.


devoncarew:
> Using the power of Git submodules, we now have a non hacky way to include flutter_web into the build.

Reviewing now; I don't have a concrete sense over whether git submodules are a good way to go, but will defer to your opinion here.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: dart
 install:
   - gem install coveralls-lcov
 dart:
-  - stable
+  - "2.3.0"
```
devoncarew: Why peg this to a specific version?

tool/grind.dart
```diff
@@ -189,7 +189,8 @@ void fuzz() {
 }
 
 @Task('Update discovery files and run all checks prior to deployment')
-@Depends(updateDockerVersion, init, discovery, analyze, test, fuzz, validateStorageArtifacts)
+@Depends(updateDockerVersion, init, discovery, analyze, test, fuzz,
```
devoncarew: How do we update the version of flutter_web in use now? Should we add a grind task to do this? Or, update our deploy docs to tell people how to do it?

.travis.yml
domesticmouse: I'm pegging this to the minimum SDK we have in our `pubspec.yaml` so we don't get trapped by testing at a `stable` that is actually higher than our min SDK version.

tool/grind.dart
domesticmouse: Adding in @kevmoo as we should update the git submodule reference on what ever the stable release cadence is for Flutter Web. I'm unsure what that cadence is, or will be. Kev?

The process for it is sketched out in [this tutorial](https://www.vogella.com/tutorials/GitSubmodules/article.html#submodules_track).

domesticmouse:
As to whether git submodules are a good idea or not, it's better than doing ad hoc checkouts on the deployment machine (which is the current process). Given that Flutter Web isn't going to be available via pub, I'm unsure what other viable approaches we have here.

tool/grind.dart
RedBrogdon: If we advance with our plan to support multiple versions of the Dart SDK, do we want to also support multiple versions of flutter_web?

domesticmouse: We have a plan to support multiple SDKs? With the current dart services backend this will require running sets of machines at each supported sdk level.

As for Flutter Web, what is the upside of supporting anything other than head? Flutter Web doesn’t have releases yet...

RedBrogdon: My comment assumes Flutter for Web will eventually adopt the stable/dev/beta/master paradigm used by the SDK, though I don't know that's guaranteed.

domesticmouse: I believe the plan is to merge it into Flutter proper. When that happens we will have to do a quite different integration, I suspect.

PR #413, opened 2019-05-30
add config.properties to .gitignore

johnpryan:


kevmoo:
No! Put this in your `.gitconfig_global` file!

Ditto with `.vscode`

https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

johnpryan:
good call. TIL...

.gitignore
```diff
@@ -21,9 +21,5 @@ dump.rdb
 # compilation artifacts
 artifacts/
 
-# editor artifacts 
-.vscode/ 
-.idea
-
 # local configuration
 config.properties
```
kevmoo: Add a newline at the end of the file, too! Makes Git happier

RedBrogdon:
> No! Put this in your `.gitconfig_global` file!
> 
> Ditto with `.vscode`
> 
> https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

If it's not in the repo's .gitignore, would that not require everyone to have a similar `.gitconfig_global` on their machine? We'd need to document that as a step in our contributor's guide, if so.

This isn't to say I think adding those files to a global ignore is a bad idea, just that it might be better to ignore it in both places.

johnpryan:
Added the editor artifacts back - it helps prevent future contributors from accidentally staging these files

kevmoo:
When the number of repos > the number of developers, I'd argue it's better
to centralize this logic on the user side.

Next week foo editor comes out and now we have to add ignores for that in
every repo where a user is using FooIDE?

johnpryan:
Good point, I'll revert that change

PR #414, opened 2019-05-30
add endpoint for assists

johnpryan:
required for dart-lang/dart-pad#1007

### DIFF
--- a/doc/generated/dartservices.dart
+++ b/doc/generated/dartservices.dart
@@ -62,6 +62,42 @@ class DartservicesApi {
     return _response.then((data) => new AnalysisResults.fromJson(data));
   }
 
+  /// Get assists for the given source code location.
+  ///
+  /// [request] - The metadata request object.
+  ///
+  /// Request parameters:
+  ///
+  /// Completes with a [AssistsResponse].
+  ///
+  /// Completes with a [commons.ApiRequestError] if the API endpoint returned an
+  /// error.
+  ///
+  /// If the used [http.Client] completes with an error when making a REST call,
+  /// this method will complete with the same error.
+  async.Future<AssistsResponse> assists(SourceRequest request) {
+    var _url = null;
+    var _queryParams = new core.Map<core.String, core.List<core.String>>();
+    var _uploadMedia = null;
+    var _uploadOptions = null;
+    var _downloadOptions = commons.DownloadOptions.Metadata;
+    var _body = null;
+
+    if (request != null) {
+      _body = convert.json.encode((request).toJson());
+    }
+
+    _url = 'assists';
+
+    var _response = _requester.request(_url, "POST",
+        body: _body,
+        queryParams: _queryParams,
+        uploadOptions: _uploadOptions,
+        uploadMedia: _uploadMedia,
+        downloadOptions: _downloadOptions);
+    return _response.then((data) => new AssistsResponse.fromJson(data));
+  }
+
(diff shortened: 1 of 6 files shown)


lib/src/api_classes.dart
```diff
@@ -232,6 +232,19 @@ class SourceEdit {
...
+class AssistsRequest {
+  @ApiProperty(required: true, description: 'The Dart source.')
+  String source;
+
+  @ApiProperty(description: 'An optional offset into the source code.')
+  int offset;
+}
+
+class AssistsResponse {
+  List<CandidateFix> assists;
```
devoncarew: `final`?

lib/src/common_server.dart
```diff
@@ -577,6 +588,22 @@ class CommonServer {
...
+    }
+    if (offset == null) {
+      throw BadRequestError('Missing parameter: \'offset\'');
+    }
+
+    await _checkPackageReferencesInitFlutterWeb(source);
+
+    Stopwatch watch = Stopwatch()..start();
+    var response = await analysisServer.getAssists(source, offset);
+    log.info('PERF: Computed fixes in ${watch.elapsedMilliseconds}ms.');
```
devoncarew: `Computed assistes...`?

lib/src/api_classes.dart
```diff
@@ -232,6 +232,19 @@ class SourceEdit {
   }
 }
 
+class AssistsRequest {
```
devoncarew: I'd add brief dartdoc here.

lib/src/api_classes.dart
```diff
@@ -232,6 +232,19 @@ class SourceEdit {
...
+  @ApiProperty(required: true, description: 'The Dart source.')
+  String source;
+
+  @ApiProperty(description: 'An optional offset into the source code.')
+  int offset;
+}
+
+class AssistsResponse {
```
devoncarew: I'd add brief dartdoc here.

lib/src/api_classes.dart
johnpryan: oops this is actually unused. removing...

doc/generated/dartservices.dart
```diff
@@ -410,6 +446,29 @@ class AnalysisResults {
   }
 }
 
+class AssistsResponse {
```
bwilkerson: In case it's useful for you to know, all of the classes needed to represent the API of the server are available in the package analysis_server_client (sdk/pkg/analysis_server_client). These classes are auto generated and are kept consistent with the specification. In particular, this class could probably be replaced by `EditGetAssistsResult`.

johnpryan: Ah! I'll make that change. Thanks!

johnpryan: @devoncarew it looks like this class isn't available in scope. I noticed that [EditGetAssistsParams](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/lib/protocol/protocol_generated.dart#L8701) isn't implemented anywhere in https://github.com/devoncarew/analysis_server_lib 

Do we need to make an update to that package or is there already an equivalent class we can use?

devoncarew: We likely want to keep these classes separate - the definition here and package: analysis_server_client. This dart library is used to generate code which implements the RESTful api for dartpad.

johnpryan: Ok, I'll leave this as-is

PR #415, opened 2019-05-31
 remove editor artifacts from .gitignore

johnpryan:
contributors should use .gitignore_global instead ([ref](https://github.com/dart-lang/dart-services/pull/413#issuecomment-497617831))

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,9 +21,5 @@ dump.rdb
 # compilation artifacts
 artifacts/
 
-# editor artifacts
-.vscode/
-.idea
-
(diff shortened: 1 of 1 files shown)


domesticmouse:
I’m voting against this change. This leads to a much poorer getting started experience by expecting developers to be aware of an obscure power user feature. Looking at GitHub’s own gitignore templates shows that ignoring editor generated files is the norm for this community, and our mantra is to meet developers where they are, not where we think they should be.

https://github.com/github/gitignore

Please revert.

PR #417, opened 2019-06-05
Rotating journal logs

domesticmouse:
Rotating and vacuuming journal log files to prevent running out of disk space on `/`.

### DIFF
--- a/lib/services_gae.dart
+++ b/lib/services_gae.dart
@@ -18,6 +18,9 @@ import 'src/flutter_web.dart';
 
 const String _API = '/api';
 
+const String _journalctlBinary = '/bin/journalctl';
+const String _sudoBinary = '/usr/bin/sudo';
+
 final Logger _logger = Logger('gae_server');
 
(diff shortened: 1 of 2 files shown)


domesticmouse:
It's the file that is growing without bound on the instances, so at this point I believe it is the cause of the outages.

domesticmouse:
And reformatted the code for line length

jonasfj:
I'm curious, what is writing logs? I just checked out staging site, and we don't seem to have processes that write log files..

domesticmouse:
It's the base OS that runs the Docker instance that houses the AppEngine Flex images, along with fluentd logging and nginx networking sidecars. It basically looks like Kubernetes + Istio, but I think it predates that tech. The logfile in question is controlled with `journalctl`, and can be easily rotated and culled, but I can't control the base image that runs AppEngine Flex. I'd much prefer if it were running Container Optimised OS instead of Debian Stretch.

PR #422, opened 2019-06-06
Adding liveness checks

domesticmouse:


lib/services_gae.dart
```diff
@@ -104,9 +106,35 @@ class GaeServer {
...
       }
-      request.response
-        ..statusCode = statusCode
-        ..close();
+      request.response.statusCode = statusCode;
+      await request.response.close();
+      return;
+    }
+
+    if (request.uri.path == _readynessCheck) {
```
devoncarew: Perhaps convert this method to a series of if-else statements, instead of using early returns? We're less likey to get an accidental fall-through to another branch then.

We could also factor this into a few different sub-methods - `_sendReadynessResponse`, `_sendLivenessResponse`, ...

lib/services_gae.dart
```diff
@@ -104,9 +106,35 @@ class GaeServer {
...
+        final file = await io.File('${tempDir.path}/livecheck.txt');
+        await file.writeAsString('testing123\n' * 1000, flush: true);
+        final stat = await file.stat();
+        if (stat.size > 10000) {
+          request.response.statusCode = io.HttpStatus.ok;
+        } else {
+          request.response.statusCode = io.HttpStatus.internalServerError;
+        }
+        await request.response.close();
+        await tempDir.delete(recursive: true);
```
jonasfj: Why not parse the output of `df` instead? hmm, I guess creating a file is a very robust way of check for disk space :)

Maybe consider removing the temporary folder in a `finally` block, so you don't risk creating more temporary files that fill up your disk, hehe,,

lib/services_gae.dart
domesticmouse: done

lib/services_gae.dart
domesticmouse: done

lib/services_gae.dart
```diff
@@ -88,61 +90,95 @@ class GaeServer {
...
-    if (request.uri.path.startsWith(_API)) {
-      if (!discoveryEnabled) {
-        apiServer.enableDiscoveryApi();
-        discoveryEnabled = true;
+  Future _processHealthRequest(io.HttpRequest request) async {
+    try {
+      final tempDir = await io.Directory.systemTemp.createTemp('healthz');
+      try {
+        final file = await io.File('${tempDir.path}/livecheck.txt');
+        await file.writeAsString('testing123\n' * 1000, flush: true);
```
RedBrogdon: TIL about multiplying strings in Dart.

domesticmouse:
@RedBrogdon The delay between tests is configurable, and I believe it is on the order of every two or three seconds. 

As to the transition from being healthy to unhealthy, by definition, as we are sampling from a continuous curve, there will come a point where the instance transitions from good to bad between samples. The point of this is to make sure a broken instance isn't left in production, continuing to serve bad traffic. This way after a couple seconds the instance will be taken out of the loop and a replacement will be being spun up.

RedBrogdon:
Got it. Thanks for the info.

PR #425, opened 2019-06-06
Upsizing disk on Flex instances

domesticmouse:


domesticmouse:
10gig, and we were running out after about a week. So this should be good for 2 weeks, by which point I believe the instances will be replaced by AppEngine.

PR #427, opened 2019-06-11
Using Platform SDK

domesticmouse:
I want to use the SDK provided in the Docker image. I believe this will make it possible to deploy from macOS as we no longer including a binary SDK in the image that we download during `grind deploy`.


lib/src/sdk_manager.dart
```diff
@@ -40,96 +37,19 @@ abstract class Sdk {
...
 }
 
-/// For this class, the cwd should be the root of the project.
-class DownloadingSdk extends Sdk {
-  static const String kSdkPathName = 'dart-sdk';
-
-  final String _versionFull;
-
-  DownloadingSdk()
-      : _versionFull = File('dart-sdk.version')
```
RedBrogdon: The `dart-sdk.version` file is used in a few other places to get the running version of the SDK. If the Docker image is now the source of truth for which version is in use, would it be a good idea to refactor that file out and use `Sdk.version` everywhere instead?

domesticmouse:
If we want the ability to use a version of the Dart SDK that isn't available as a Docker image, given the Docker image is rolled as part of the release process that also creates the Dart downloads that the old process uses, then we have to make the launch process even more technically demanding, and currently it is already so complex I'm the only bunny who does launches successfully. 

I like the idea of reducing the complexity of the deploy as it makes it less special snowflake. I dislike being the only person that can publish services, as currently the bus risk factor way too low.

lib/src/sdk_manager.dart
domesticmouse: Yup, I'm happy to remove `dart-sdk.version` and refactor the code base to use `Sdk.version`.

RedBrogdon:
SGTM.

lib/src/sdk_manager.dart
domesticmouse: Done

PR #429, opened 2019-06-14
First pass at supporting package:test

domesticmouse:
This is a first pass at whitelisting `package:test` to make authoring codelabs easier. This is a fairly major change to our compilation strategy and looks like it might impact compilation speed, because we will now be running pub at the start of each compilation. I'd appreciate feedback on this approach.

### DIFF
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -140,7 +140,7 @@ environment:
   sdk: '>=${SdkManager.sdk.version} <3.0.0'
 
 dependencies:
-  test: any
+  matcher: any
 ''');
 
       final arguments = <String>[
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -129,23 +129,31 @@ class Compiler {
     Directory temp = Directory.systemTemp.createTempSync('dartpad');
 
     try {
-      List<String> arguments = <String>[
-        '--modules=amd',
-      ];
-
-      if (flutterWebManager.usesFlutterWeb(imports)) {
-        arguments.addAll(<String>['-s', flutterWebManager.summaryFilePath]);
-      }
-
       String compileTarget = path.join(temp.path, kMainDart);
       File mainDart = File(compileTarget);
       mainDart.writeAsStringSync(input);
+      File pubspec = File(path.join(temp.path, 'pubspec.yaml'));
+      pubspec.writeAsStringSync('''
+name: dartpad_main
 
-      arguments.addAll(<String>['-o', path.join(temp.path, '$kMainDart.js')]);
-      arguments.add('--single-out-file');
-      arguments.addAll(<String>['--module-name', 'dartpad_main']);
-      arguments.add(compileTarget);
-      arguments.addAll(<String>['--library-root', temp.path]);
+environment:
+  sdk: '>=2.3.2 <3.0.0'
+
+dependencies:
+  test: ^1.6.4
+''');
+
+      final arguments = <String>[
+        '--modules=amd',
+        if (flutterWebManager.usesFlutterWeb(imports)) ...[
+          '-s',
+          flutterWebManager.summaryFilePath
+        ],
+        ...['-o', path.join(temp.path, '$kMainDart.js')],
+        '--single-out-file',
+        compileTarget,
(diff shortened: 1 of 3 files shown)


domesticmouse:
This is a fix for https://github.com/dart-lang/dart-pad/issues/1120

devoncarew:
I had an older PR (since landed and mostly reverted) that performed some of what we'd need: https://github.com/dart-lang/dart-services/pull/136; https://github.com/dart-lang/dart-services/pull/372.

The technique we had for general package support (a super set of supporting package:test, so perhaps unnecessarily general), was:

- for a request, scan the source for import statements (the code to do this still exists in the repo)
- create a pubspec file with `dependencies:` for the package imports found, using the version constraint `any`
- use `pub get` to resolve the necessary package versions
- cache that resolved info in memcache / redis
- create our own, local, pub cache directory
- populate it with the package versions resolved above, with each package version downloaded as a tar.gz from the pub site
- manually create a `.packages` file using the resolved version info, with the package locations pointing into the local package cache
- run compile / analyze commands using the created .packages file

I'm not sure how this would work wrt package summaries and compiling with DDC. The above would all be fairly performant - it would amortize to pretty low overhead per call. The `pub get` call would be cached, so most requests would be serviced from the cache - the time for that would drop to ~0. As an individual server ran, it would progressively fill its local pub cache, so most requests would not need to download packages.

Happy to VC if you'd like more details.

domesticmouse:
PTAL @RedBrogdon 

I'd like to land this so you can experiment with your proposed matcher based testing scripts on DartPad.

lib/src/compiler.dart
```diff
@@ -129,23 +129,32 @@ class Compiler {
...
-
-      if (flutterWebManager.usesFlutterWeb(imports)) {
-        arguments.addAll(<String>['-s', flutterWebManager.summaryFilePath]);
-      }
-
       String compileTarget = path.join(temp.path, kMainDart);
       File mainDart = File(compileTarget);
       mainDart.writeAsStringSync(input);
+      File pubspec = File(path.join(temp.path, 'pubspec.yaml'));
+      pubspec.writeAsStringSync('''
```
devoncarew: This PR doesn't run `pub get` after creating the pubspec file, or handle cases where we instead want to use `.packages` info from flutter web projects. We'd also need to whitelist use of `package:matcher` / `package:test`? `pub get` is not super performant, so I think we'd want to run it once and cache the resulting info rather than run it per compile or analysis call.

I think we're better off with extending (and possibly renaming) the `FlutterWebManager` class.

domesticmouse: I understand your desire to do the right thing here, engineering wise, and merge the `CompileDDC` and `Compile` codepaths into a new `CompilationManager` that appropriately deals with Flutter Web, Angular Web, and plain dart code. 

However the above is going to take a fair amount of time, and given my commitments to presenting through South East Asia in July and in China in September, I'm unsure if it will get it done before mid Q4.

This patch, as it stands, is enough to give us validation for the work that @RedBrogdon and @legalcodes want to do with testing infrastructure. I'm happy to take an action item to build `CompilationManager` as time permits. 

We can even roll this PR back once @RedBrogdon and @legalcodes are happy with their testing, but I have a strong suspicion that the cost of extra CPU required by the current implementation approach is going to be completely swamped by the engineering cost of pushing back implementation until late Q4.

devoncarew: I don't have any objections if you'd like to land this to unblock some investigation work elsewhere. I just don't think that we should plan to productionize this (given that there are other techniques that will have better performance and other characteristics).
 
Given that this patch doesn't run `pub get` after creating the pubspec.yaml file, I'm not sure this code does?

RedBrogdon: It looks like I can get everything I need to test in the short term simply by turning on flutter_web and inlining the package code:

https://dartpad.dartlang.org/experimental/embed-new.html?fw=true&id=fa8cbc5ee5d710f5ada762d97d5d4237

That's plenty for me right now, and if we're going to do something hacky to enable me to futz about with the test code, I'd much prefer it be a blob of ugly code pasted into a gist than anything on the back end.

PR #432, opened 2019-06-27
Caching compileDDC results in Redis

domesticmouse:
Fix for https://github.com/dart-lang/dart-pad/issues/1126

### DIFF
--- a/lib/src/common_server.dart
+++ b/lib/src/common_server.dart
@@ -441,14 +441,14 @@ class CommonServer {
 
     await _checkPackageReferencesInitFlutterWeb(source);
 
-    String sourceHash = _hashSource(source);
-    String memCacheKey = '%%COMPILE:v0'
+    final sourceHash = _hashSource(source);
+    final memCacheKey = '%%COMPILE:v0'
         ':returnSourceMap:$returnSourceMap:source:$sourceHash';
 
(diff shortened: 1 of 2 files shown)


lib/src/common_server.dart
```diff
@@ -441,44 +441,44 @@ class CommonServer {
...
     await _checkPackageReferencesInitFlutterWeb(source);
 
-    String sourceHash = _hashSource(source);
-    String memCacheKey = '%%COMPILE:v0'
+    final sourceHash = _hashSource(source);
+    final memCacheKey = '%%COMPILE:v0'
         ':returnSourceMap:$returnSourceMap:source:$sourceHash';
 
-    final String result = await checkCache(memCacheKey);
+    final result = await checkCache(memCacheKey);
```
devoncarew: As a drive by comment - does the cache impl. automatically add the sdk version to cache keys? If not, we may consider doing that.

domesticmouse: Add Dart SDK Version to `RedisCache`'s `_genKey`.

domesticmouse:
TBR: @RedBrogdon 

Forcing this to bring the Dart SDK version aware caching into production. In theory Dart 2.3 and Dart 2.4 shouldn't have user visible changes, but I'd rather our users not trip over it.

PR #438, opened 2019-07-08
Making compilation properly async

domesticmouse:
Hey Devon, I suspect the 500s we see under load are due to the compilation back end going out to lunch while it compiles things. How's this fix feel to you?

### DIFF
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -63,7 +63,7 @@ class Compiler {
       ]);
     }
 
-    Directory temp = Directory.systemTemp.createTempSync('dartpad');
+    Directory temp = await Directory.systemTemp.createTemp('dartpad');
 
     try {
       List<String> arguments = <String>[
@@ -78,7 +78,7 @@ class Compiler {
 
       String compileTarget = path.join(temp.path, kMainDart);
       File mainDart = File(compileTarget);
-      mainDart.writeAsStringSync(input);
+      await mainDart.writeAsString(input);
 
       File mainJs = File(path.join(temp.path, '$kMainDart.js'));
       File mainSourceMap = File(path.join(temp.path, '$kMainDart.js.map'));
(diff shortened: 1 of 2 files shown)


lib/src/compiler.dart
```diff
@@ -63,7 +63,7 @@ class Compiler {
       ]);
     }
 
-    Directory temp = Directory.systemTemp.createTempSync('dartpad');
+    Directory temp = await Directory.systemTemp.createTemp('dartpad');
```
devoncarew: I think in general we want to keep the `fooSync` version of most of the dart:io calls (`existsSync`, `readAsStringSync`, ...).

lib/src/flutter_web.dart
```diff
@@ -132,10 +132,10 @@ $_samplePackageName:lib/
     return null;
   }
 
-  void _runPubGet() {
+  Future<void> _runPubGet() async {
     _logger.info('running pub get (${_projectDirectory.path})');
 
-    ProcessResult result = Process.runSync(
+    ProcessResult result = await Process.run(
```
devoncarew: Perhaps, keep the change to make the `pub get` call async, but change to a version of process exec where you can log stdout as pub get makes progress?

lib/src/compiler.dart
domesticmouse: What is the advantage of using the `*Sync` versions here? We effectively stop the server from responding for the length of the pub package retrieval and compilation, which takes a second or so, going by the logs.

domesticmouse:
The `cla: no` flag flip is due to syncing to head to keep the PR up to date.

lib/src/compiler.dart
devoncarew: Most of these sync methods will execute on the order of a millisecond. 

> I suspect the 500s we see under load are due to the compilation back end going out to lunch while it compiles things

Are you worried about the server not responding to other requests while processing the current one? We're only sent one at a time I believe. Are ew not responding to health requests? It would have thought that a 500 response was due to the server throwing an exception.

The only thing that's taking time here is the synchronous `pub get` call - totally makes sense to cvt that to async. Happy to IM as well, to better understand the problem.

domesticmouse: Since I started soak testing this PR on production traffic, instances of 500 responses have practically disappeared from our dashboard summary. Your understanding of requests coming at the server one at a time aren't correct btw, it just looks like it because we serve requests that take about a second to complete and we have slightly more servers than the average QPS that we are serving. Under load we start serving multiple requests in parallel, which would explain why we see spikes in 500s during high load periods.

devoncarew: I'm at somewhat of a loss to explain why this PR would help with the 500s...

PR #439, opened 2019-07-23
Update flutter web version

domesticmouse:
TBR @johnpryan


domesticmouse:
This is now running in prod

PR #442, opened 2019-08-16
Send cursor position and linked edit groups down with quick fixes

RedBrogdon:


RedBrogdon:
q.v. https://github.com/dart-lang/dart-pad/pull/1210

lib/src/api_classes.dart
```diff
@@ -7,6 +7,7 @@ library services.api_classes;
 
 import 'dart:convert';
 
+import 'package:analysis_server_lib/analysis_server_lib.dart';
```
johnpryan: I think we probably want to avoid using types from the analysis server API in the RESTful API for DartPad ([ref](https://github.com/dart-lang/dart-services/pull/414#discussion_r289186449)

domesticmouse: I'm curious what is being used from this import. `LinkedEditGroup` is the only new type in this delta, and that class is defined in `dartservices.dart`, yes?

RedBrogdon: @johnpryan I was looking at this (admittedly way old) comment from Luke:

https://github.com/dart-lang/dart-services/blob/master/lib/src/api_classes.dart#L174

Which led me to believe that exposing the class directly was the desired approach. I'd prefer the API and the analyzer be less tightly coupled, so (since it seems like we're in agreement on that) I'll refactor.

@domesticmouse LinkedEditGroup, LinkedEditSuggestion, and Position all come out of that file. `dartservices.json` is then generated from those definitions by `grind discovery`.

RedBrogdon:
Decoupled analyzer and API, tested with both old and new versions of dart-pad.

domesticmouse:
This is now deployed in production, so you should be able to deploy the front-end at your leisure. =)

PR #449, opened 2019-09-11
Updates for 2.5

RedBrogdon:
Includes the following changes:

* Updates SDK/image to 2.5.0
* Pins `_discoveryapis_common` dependency to 0.1.8+1
  - This works around the CORS issue with 0.1.9
* Updates the `rpc` dep to point to the 0.6.2 code, which hasn't yet been published
  - This fixes the break when rpc is run on Dart 2.5.0.

Once merged, dart-services can be deployed and we'll have a working 2.5.0 backend.


pubspec.yaml
```diff
@@ -3,10 +3,10 @@ publish_to: none
 
 environment:
   # Keep min SDK aligned with .travis.yml and FROM line of Dockerfile
-  sdk: '>=2.4.1 <3.0.0'
+  sdk: '>=2.5.0 <3.0.0'
 
 dependencies:
-  _discoveryapis_commons: ^0.1.8
+  _discoveryapis_commons: 0.1.8+1
```
johnpryan: nit: ">=0.1.8 <0.1.9" might be better in the case where we need to make updates to this version

RedBrogdon: Not sure I follow. Do you mean publishing another version of the package in between 0.1.8 and 0.1.9?

I'd prefer to pin this for now if only because I know I've manually tested dart-services and dart-pad with specific versions and can guarantee they're working together, which allows us to ship 2.5.

We will still need to determine why 0.1.9 isn't working for us, which will result in another change to this dependency.

johnpryan: 0.1.8+2 could be published, right?

RedBrogdon: Discussed offline. Will merge this and restore the range once 0.1.9 is figured out.

PR #451, opened 2019-09-16
Update Discovery APIs Commons to 0.1.9

domesticmouse:
Running as http://20190917t075845-dot-dart-services.appspot.com/

/cc @johnpryan

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -7,7 +7,7 @@ packages:
       name: _discoveryapis_commons
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "0.1.8+1"
+    version: "0.1.9"
   analysis_server_lib:
     dependency: "direct main"
     description:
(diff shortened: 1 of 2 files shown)


johnpryan:
Looks good to me. It looks like we still need to keep the frontend package at 0.1.8+2 though...  I verified it works locally with _discoveryapis_commons 0.1.8+2, but I wasn't able to get it working with 0.1.9:

```
Access to XMLHttpRequest at 'http://20190917t075845-dot-dart-services.appspot.com/api/dartservices/v1/compileDDC?alt=json' from origin 'http://localhost:8000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

domesticmouse:
Can/should we just add `Access-Control-Allow-Origin: *` to our services API?

PR #456, opened 2019-10-11
Test with prod and dev

domesticmouse:


domesticmouse:
Yup, it's preparation work for monorepo-izing. I'm setting it up so that we can run it under different versions of Dart for Flutter.

PR #465, opened 2019-11-06
Unforks Flutter SDK compilation and analysis

RedBrogdon:

PR #467, opened 2019-11-13
Updates compilation and analysis to use an unforked version of the Flutter SDK

RedBrogdon:
* Replaces flutter_web submodule with flutter/flutter
* Updates Dockerfile to clone flutter/flutter
* Updates the compiler to use DDK and the unforked Flutter SDK for Flutter.
* Updates common_server to spin up two analysis server instances:
  - One uses the platform Dart SDK for Dart code
  - One uses Flutter's cached Dart SDK for Flutter code.
* Adds tests to verify the above.

This PR is intended to replace #465.


lib/src/flutter_web.dart
```diff
@@ -71,26 +75,29 @@ $_samplePackageName:lib/
 
     await _runPubGet();
 
-    final String sdkVersion = SdkManager.sdk.version;
+    final String sdkVersion = flutterSdk.versionFull;
 
-    // download and save the flutter_web.sum file
+    // Download and save the flutter_web.dill file.
```
domesticmouse: Raise a task against me to replace this download with a direct dependency after this PR lands.

.dockerignore
```diff
@@ -0,0 +1,4 @@
+# Docker should ignore the local copy of Flutter and install its own,
+# since the Docker container is likely running a different OS than the
+# dev's machine.
```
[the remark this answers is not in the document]
RedBrogdon: Good point. Fixed.

app.yaml
```diff
@@ -27,3 +27,4 @@ liveness_check:
 
 readiness_check:
   path: /_ah/ready
+  app_start_timeout_sec: 600
```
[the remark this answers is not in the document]
RedBrogdon: Glad you caught this. Before giving up and switching to a fresh SDK install via Dockerfile, I thought my failing instances just needed more time to spin up. I'll take it out.

Dockerfile
```diff
@@ -16,12 +20,37 @@ RUN pub get --offline
...
+RUN cat flutter/bin/cache/dart-sdk/version
+
 EXPOSE 8080 8181 5858
 
 # Clear out any arguments the base images might have set and ensure we start
 # the Dart app using custom script enabling debug modes.
 CMD []
+
+# Switch back to root to run the application.
+USER root
```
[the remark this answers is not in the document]
RedBrogdon: Sent via email, since it's too big to share here.

lib/src/flutter_web.dart
```diff
@@ -71,26 +75,29 @@ $_samplePackageName:lib/
...
   String get summaryFilePath {
-    return path.join(_projectDirectory.path, 'flutter_web.sum');
+    return path.join(_projectDirectory.path, 'flutter_web.dill');
   }
 
   static final Set<String> _flutterWebImportPrefixes = <String>{
-    'package:flutter_web',
-    'package:flutter_web_ui',
-    'package:flutter_web_test',
+    'package:flutter',
```
johnpryan: have you tested `dart:io` yet?

Dockerfile
```diff
@@ -1,12 +1,16 @@
 # Keep aligned with min SDK in pubspec.yaml and Dart test version in .travis.yml
 FROM google/dart:2.6.0
 
+# The specific commit that dart-services should use. This should be kept
+# in sync with the flutter submodule in the dart-services repo.
+# (run `git rev-parse HEAD` from the flutter submodule to retrieve this value.
+ARG FLUTTER_COMMIT=fbabb264e0ab3e090d6ec056e0744aaeb1586735
```
johnpryan: I would like to look into parsing this from the submodule, but we can circle back to that.

RedBrogdon: SGTM

lib/src/flutter_web.dart
RedBrogdon: You can import it, which is a problem, since we need to correctly tell people it's not going to work in a browser. There's an open issue for this @ https://github.com/dart-lang/dart-pad/issues/857

johnpryan: oh sorry, I meant `dart:ui`

PR #469, opened 2019-11-28
Land current code for nullsafety.dartpad.dev into nnbd branch

RedBrogdon:


RedBrogdon:
CC @devoncarew

lib/src/sdk_manager.dart
```diff
@@ -61,7 +61,7 @@ class PlatformSdk extends Sdk {
   String get versionFull => _versionFull;
 
   @override
-  String get sdkPath => path.dirname(path.dirname(Platform.resolvedExecutable));
+  String get sdkPath => path.join(Directory.current.path, 'dart-sdk');
```
johnpryan: What's the reason for this change? It might be subject to break if the SDK is installed somewhere else (like on my mac)?

test/analysis_server_test.dart
```diff
@@ -61,28 +61,28 @@ void main() { foo() }
...
 void main() {
   int a = <String, int>{'a': 1}['a']!;
-  print(a == null ? 'null' : 'not null');
+  print(a);
 }
 ''';
 
 const sampleNNBDCoreLibraryFailureCode = '''
 void main() {
   int a = <String, int>{'a': 1}['a'];
```
domesticmouse: I'm curious what the difference between `sampleNNBDCoreLibraryCode` and `sampleNNBDCoreLibraryFailureCode`

RedBrogdon: An exclamation point.

domesticmouse: It really hides in there, doesn't it?

RedBrogdon: Probably a good indicator I should throw a comment on it.

domesticmouse: Sgtm =)

RedBrogdon:
This PR is still a throwaway. I don't intend to land it, and am keeping it here mainly as a way to compare master and the null safety changes.

domesticmouse:
Converted to draft status to reflect the non-goal of landing this PR.

RedBrogdon:
> Converted to draft status to reflect the non-goal of landing this PR.

Thanks. I know realize we don't have the WIP checker on this repo. We should set that up.

lib/src/sdk_manager.dart
RedBrogdon: (I'm sure we've talked about this in between now and when it was posted, but I'd still feel guilty closing out the comment without responding)

At the time this code was written, it was necessary to do a custom build of the SDK in order to get access to null safety. That's no longer true, but we still need one for *sound null safety*. 

As a result, I had to add a step to the Dockerfile that downloads my custom build of the Dart SDK from cloud storage while creating the image. This class represents that SDK (or a local SDK in the same folder, when building/testing locally).

RedBrogdon:
@domesticmouse Mind taking another look at this before it lands?

PR #473, opened 2019-12-20
Updates flutter commit to latest dev channel

RedBrogdon:
* Rolls new Flutter commit.
* Updates some tests to account for new analyzer responses.
* Works around the lack of `--single-out-file` in the new version of the Dart SDK.
* Adds a new grind task that blows away the Flutter submodule, recreates, and re-initalizes it. I had some trouble with weird internal state in the SDK submodule, and created this task as a way to "clean" the Flutter submodule.


RedBrogdon:
Note: This change should/will not be deployed to production until after the holidays.

RedBrogdon:
CC @sigmundch.

Thanks again for the advice!

PR #475, opened 2020-01-08
Pedantic 1.9

domesticmouse:
Updating to pedantic v1.9, and moving to a final preferred status.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,4 +1,4 @@
-include: package:pedantic/analysis_options.1.8.0.yaml
+include: package:pedantic/analysis_options.1.9.0.yaml
 
 analyzer:
   strong-mode:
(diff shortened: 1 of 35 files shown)


analysis_options.yaml
```diff
@@ -8,3 +8,9 @@ analyzer:
     - 'third_party/**'
     - 'doc/generated/**'
     - 'flutter/**'
+
+linter:
+  rules:
+    - prefer_final_fields
+    - prefer_final_in_for_each
+    - prefer_final_locals
```
RedBrogdon: @johnpryan may have thoughts on this one. I like mandating use of inferred types, but might not want to stomp on anyone using `var` instead of `final` for a local.

johnpryan: I'm not going to argue :)

PR #476, opened 2020-01-31
Update flutter and dart

domesticmouse:


domesticmouse:
After deploying this version we will need to update samples as they wind up showing deprecation notices for Material theme changes. 

/cc @johnpryan for https://gist.github.com/johnpryan/b6409e10de32b280b8938aa75364fa7b which will need `s/display1/heading4/`. 

Screen Shot 2020-01-31 at 1 13 28 pm

domesticmouse:
SGTM. Monday may be hazy due to jet lag, but it only requires a traffic switch to push live.

domesticmouse:
Actually, silly question time, I upgraded flutter to latest master. That’s possibly the wrong branch to track. Should we be tracking `dev` or `beta` instead?

johnpryan:
I'm fine with master since is has the latest web tooling. We just need to do some sanity checks to make sure we're not *too* bleeding edge.

PR #478, opened 2020-02-10
Dropping unused _dartpadsupportservicesApi

domesticmouse:
Depends on https://github.com/dart-lang/dart-pad/pull/1478


tool/grind.dart
```diff
@@ -17,6 +17,14 @@ import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as path;
 
 Future<void> main(List<String> args) async {
+
+  // run flutter doctor to make sure flutter has downloaded artifacts
+  // required for compilation
+  run(
```
RedBrogdon: If you're using the setupFlutterSubmodule task, this should get done for you.

(not that you'd want this task to depend on that one, but if you're using that task to set up your submodule, it shouldn't ever be necessary to call the doctor here).

domesticmouse: Nope, doesn't work. Making `deploy` depend on `setupFlutterSubmodule` is insufficient:

```dart
grind deploy
Unhandled exception:
FileSystemException: Cannot open file, path = '/Users/brettmorgan/Documents/GitHub/dart-services/flutter/bin/cache/dart-sdk/version' (OS Error: No such file or directory, errno = 2)
#0      _File.open.<anonymous closure> (dart:io/file_impl.dart:366:9)
#1      _RootZone.runUnary (dart:async/zone.dart:1381:54)
#2      _FutureListener.handleValue (dart:async/future_impl.dart:139:18)
#3      Future._propagateToListeners.handleValueCallback (dart:async/future_impl.dart:680:45)
#4      Future._propagateToListeners (dart:async/future_impl.dart:709:32)
#5      Future._completeWithValue (dart:async/future_impl.dart:524:5)
#6      Future._asyncComplete.<anonymous closure> (dart:async/future_impl.dart:554:7)
#7      _microtaskLoop (dart:async/schedule_microtask.dart:43:21)
#8      _startMicrotaskLoop (dart:async/schedule_microtask.dart:52:5)
#9      _runPendingImmediateCallback (dart:isolate-patch/isolate_patch.dart:118:13)
#10     _RawReceivePortImpl._handleMessage (dart:isolate-patch/isolate_patch.dart:175:5)
```

RedBrogdon: That's weird. Step 2 in setupFlutterSubmodule is to run `flutter doctor`. We should check in about it tomorrow morning, because if it's not working, setupFlutterSubmodule needs to be fixed.

domesticmouse: My belief is that the trigger here is that a fresh checkout of the codebase is insufficient for the `await SdkManager.flutterSdk.init();` in grind.dart's `main()`. Thus my original fix of putting a `flutter doctor` exec where I had it, to force the flutter checkout into a known sane state.

PR #481, opened 2020-02-12
Split common server

domesticmouse:
In preparation for replacing the RPC package, turn `CommonServer` into a proxy wrapper for the real implementation. Further refactoring will be required once we introduce the new API interface.


lib/src/common_server.dart
```diff
@@ -5,485 +5,130 @@
...
-    flutterAnalysisServer = AnalysisServerWrapper(
-        flutterWebManager.flutterSdk.sdkPath, flutterWebManager);
-
-    compiler =
-        Compiler(SdkManager.sdk, SdkManager.flutterSdk, flutterWebManager);
+    String sdkPath,
+    FlutterWebManager flutterWebManager,
+    ServerContainer container,
+    ServerCache cache,
+  ) : _impl = CommonServerImpl(sdkPath, flutterWebManager, container, cache);
```
johnpryan: You could also make this class [abstract with a factory constructor](https://dartpad.dev/796fdd21647fb32663c35c7c3cb0cd19)

domesticmouse: The intent is that I will be introducing a second version of `CommonServer` that serves Protos over HTTPS using `CommonServerImpl` as the implementation, on the way to eventually deleting `CommonServer` once we have verified that usage of the old RPC based API has dropped sufficiently. I really don't want `CommonServerImpl` being tied in anyway to `CommonServer` as that blocks the migration path.

I probably need better naming for these pieces of the sliding jigsaw puzzle.

domesticmouse:
I've mostly merged the changes from #482, but I haven't moved the location of `_running` from `warmup` to `init` as that feels like a mistake. Happy to implement once someone explains why this change was made.

domesticmouse:
I moved the location of setting `_running` to true from `warmup` to `init` to conform with the current reality.

softmarshmallow:
We need README update in order to run this version. 
I couldnt find what to pass to proxy-target for local dev serving

domesticmouse:
What are you attempting to test? If you are doing local development and wanting to do local compilations, you don't need to pass anything.

If you want your local server to just proxy requests to another dart-services instance, then you pass the base URL of that server as the proxy.

Please raise a new issue and I'll do the fix.

PR #482, opened 2020-02-14
Updated logging, ready checks, warmup script

RedBrogdon:
* Adds stacktrace logging for >= SEVERE messages.
* Tweaks the warmup script so it's useful for sharded AEF projects.
* Updates readiness check to use the new `running` flag on CommonServer


lib/src/common_server.dart
```diff
@@ -91,14 +93,15 @@ class CommonServer {
         exit(code);
       }
     }));
+
+    _running = true;
```
domesticmouse: Why did you move the setting of `_running` to true from `warmup` to here?

PR #484, opened 2020-02-19
Dart Services API v2

domesticmouse:
This PR adds a new API version, utilising Protobuf to define the wire format content. It exposes both  Protobuf and JSON wire formats. The content of the Protobuf API has been setup to match the current v1 API as closely as possible.

This PR includes:
 - Adds the protobuf endpoints
   - `api/dartservices/v2/analyze`
   - `api/dartservices/v2/compile`
   - `api/dartservices/v2/complete`
   - `api/dartservices/v2/document `
   - `api/dartservices/v2/fixes`
   - `api/dartservices/v2/version`
 - Updated `example` app, upgraded to Dart 2 and usable with `webdev`, for exploring the new API
 - Refactored test suite, pulling out the Redis tests, and adding matching coverage for v1 and v2 of the API

The output from the new API v2 endpoints mostly match the similarly named endpoints under `api/dartservices/v1/`. The primary difference in output is that this version doesn't return empty collections. This is most visible with the analyze endpoint, where v1 will return the following structure for an empty analysis:

```json
{
   "issues":[
   ],
   "packageImports":[
   ]
}
```

The new v2 API will instead return `{}`.


lib/services_dev.dart
```diff
@@ -103,15 +110,23 @@ class EndpointsServer {
   ApiServer apiServer;
   bool discoveryEnabled;
   CommonServer commonServer;
+  CommonServerImpl commonServerImpl;
```
johnpryan: Does this need to be field on the instance? Consider leaving this as a local variable in the constructor.

test/common_server_test.dart
```diff
@@ -328,7 +114,7 @@ void defineTests() {
     });
 
     tearDownAll(() async {
-      await server.shutdown();
+      await commonServerImpl.shutdown();
```
johnpryan: I would expect server.shutdown() to shut down the `CommonServerImpl` object it's storing. Consider making `server.shutdown()` do this instead.

tool/fuzz_driver.dart
```diff
@@ -115,7 +117,7 @@ Usage: slow_test path_to_test_collection
   print('Shutting down');
 
   await analysisServer.shutdown();
-  await server.shutdown();
+  await commonServerImpl.shutdown();
```
johnpryan: Same here, I would expect `server.shutdown()` to take care of the shutting down the `CommonServerImpl` object

test/common_server_test.dart
domesticmouse: Shutdown isn't part of the API exposed by either the API Server or the Protobuf API. 

I think you are confusing references and ownership.

tool/fuzz_driver.dart
domesticmouse: Shutdown isn't part of the exposed API, so no.

lib/services_dev.dart
domesticmouse: Done

tool/fuzz_driver.dart
johnpryan: That's because it was taken out though, right?

test/common_server_test.dart
johnpryan: I think my question is more around why CommonServer's API needed to be changed? I was expecting CommonServer's API to stay the same, since it's the API wrapper. Is the idea to eventually remove CommonServer altogether?

domesticmouse: `CommonServer` was a messy combination of interface and implementation. 

To make it possible to expose the API on both `RPC` and `Protobuf` I needed to separate the interface from implementation, then duplicate the interface part because the types are incompatible between `RPC` and `Protobuf` land. specifically the thrown exceptions.

`CommonServer` will go away when we retire the `RPC` api surface. As to when that will be possible, I have no idea. We will see it in the usage statistics.

tool/fuzz_driver.dart
domesticmouse: I separated the interface and the implementation of the API. The shutdown call was never exposed through `RPC`, so it doesn't belong in the API surface.

domesticmouse:
As to `Protobuf` using the dart toolchain from inside your included flutter repo, that's definitely broken. What does `which dart` from your nearest shell show?

README.md
```diff
@@ -13,11 +13,33 @@ It provides many of DartPad's features, including static analysis (errors and wa
 compilation to JavaScript, code completion, dartdoc information, code formatting, and
 quick fixes for issues.
 
+## Getting set up
```
domesticmouse: Hey @johnpryan, how do these instructions look for covering new Protobuf deps?

test/common_server_test.dart
johnpryan: Makes sense, thanks.

tool/fuzz_driver.dart
johnpryan: Makes sense now, thanks.

README.md
johnpryan: LGTM

lib/src/common_server_proto.dart
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  Router get router => _$CommonServerProtoRouter(this);
+
+  // We are serving requests that are arriving in both Protobuf binary encoding,
+  // and Protobuf JSON encoding. To handle this we need the ability to decode
+  // the requests and encode the responses. We also need to know how to do the
+  // work the request is requesting.
+
+  Future<Response> _serve<I, O extends GeneratedMessage>(
```
RedBrogdon: This is a naming nit, consider renaming this to "processRequest" or something similar. I was confused at first about exactly what the method did.

protos/dart_services.proto
```diff
@@ -0,0 +1,177 @@
+syntax = "proto3";
```
RedBrogdon: nit: Copyright header.

tool/grind.dart
```diff
@@ -272,3 +272,18 @@ void discovery() {
     '--output-dir=doc/generated'
   ]);
 }
+
+@Task('Generate Protobuf classes')
+void protobuf() {
```
RedBrogdon: nit: `generateProtos` or something similar?

README.md
```diff
@@ -13,11 +13,33 @@ It provides many of DartPad's features, including static analysis (errors and wa
...
+This project is built with [grinder](https://pub.dev/packages/grinder). To install, please run:
+
+```bash
+pub global activate grinder
+```
+
+The dart-services v2 API is defined in terms of Protobuf, so we need to install the Protobuf `protoc` compiler. For Mac:
+
+```bash
+brew install protobuf
```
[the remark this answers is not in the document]
domesticmouse: For Windows, I have NFI. I'll add notes for Ubuntu that basically replicate the steps from test setup

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "git.ignoreLimitWarning": true
```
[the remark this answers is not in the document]
domesticmouse: Deleted

README.md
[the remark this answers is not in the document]
domesticmouse: Actually, I'm deferring to the official Protobuf installation instructions.

lib/src/common_server_proto.dart
domesticmouse: Done

protos/dart_services.proto
domesticmouse: Done

tool/grind.dart
domesticmouse: Done

README.md
```diff
@@ -13,11 +13,33 @@ It provides many of DartPad's features, including static analysis (errors and wa
...
 
+## Getting set up
+
+This project is built with [grinder](https://pub.dev/packages/grinder). To install, please run:
+
+```bash
+pub global activate grinder
+```
+
+The dart-services v2 API is defined in terms of Protobuf, so we need to install the Protobuf `protoc` compiler. For Mac:
```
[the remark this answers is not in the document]
domesticmouse: Took out the we's.

example/web/apitest.dart
```diff
@@ -0,0 +1,209 @@
...
+import 'dart:html';
+import 'package:codemirror/codemirror.dart';
+import 'services_utils.dart' as utils;
+
+const versions = ['v1', 'v2'];
+
+utils.SanitizingBrowserClient client;
+
+void main() {
+  setupAnalyze();
```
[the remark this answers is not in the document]
domesticmouse: Added assists and compileDDC to the example web app.

README.md
```diff
@@ -13,11 +13,33 @@ It provides many of DartPad's features, including static analysis (errors and wa
...
+
+This project is built with [grinder](https://pub.dev/packages/grinder). To install, please run:
+
+```bash
+pub global activate grinder
+```
+
+The dart-services v2 API is defined in terms of Protobuf, which requires
+the installation of the Protobuf `protoc` compiler. Please see 
+[Protocol Buffers]()https://developers.google.com/protocol-buffers/)
```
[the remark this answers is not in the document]
domesticmouse: Dropped

lib/src/protos/dart_services.pbjson.dart
```diff
@@ -5,28 +5,28 @@
...
+  '1': 'SourceRequest',
   '2': const [
     const {'1': 'source', '3': 1, '4': 1, '5': 9, '10': 'source'},
     const {'1': 'offset', '3': 2, '4': 1, '5': 5, '10': 'offset'},
   ],
 };
 
-const AnalyzeReply$json = const {
-  '1': 'AnalyzeReply',
+const AnalysisResults$json = const {
```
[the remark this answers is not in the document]
domesticmouse: We are aligned between DartPad and dart-services, but we are not internally consistent inside the proto definition. I'd like to rename things for consistency in a followup pair of PRs, one for `dart-services` and one for `dart-pad`.

kevmoo:
Is it faster? 😃

domesticmouse:
@kevmoo It's step 1 of a multi-step process for weening ourselves off of `package:rpc`, which should make @mit-mit lots happier =) (Step 2 is https://github.com/dart-lang/dart-pad/pull/1501)

But no, it's not appreciably faster, as we are still sending JSON blobs back and forth for the time being. Even migrating from JSON to Protobuf for the wire format wouldn't buy us appreciable speed up, `package:protobuf` doesn't appear to compress String content from the network traces I've seen.

kevmoo:
Cool. Just checking. Ditching pkg:rpc is good enough reason. :-)

PR #487, opened 2020-03-17
Expose Flutter SDK version through API

domesticmouse:


lib/src/common_server_proto.dart
```diff
@@ -255,9 +256,11 @@ class CommonServerProto {
       ..sdkVersionFull = apiResponse.sdkVersionFull
       ..runtimeVersion = apiResponse.runtimeVersion
       ..appEngineVersion = apiResponse.appEngineVersion
-      ..servicesVersion = apiResponse.servicesVersion;
+      ..servicesVersion = apiResponse.servicesVersion
+      ..flutterVersion = apiResponse.flutterVersion;
   }
 
+
```
RedBrogdon: nit: whitespace

domesticmouse: fixed, via `dartfmt`

domesticmouse:
As for testing, I haven't tested with the existing DartPad. However, looking at the code that Discovery generates for the API, it'll have no issue ignoring additional fields. It's looking for specific named fields in the json parsed object tree.

tool/grind.dart
```diff
@@ -274,16 +269,47 @@ void discovery() {
...
-  print(result.stdout);
-  if (result.exitCode != 0) {
-    throw 'Error generating the Protobuf classes\n${result.stderr}';
-  }
 
   // generate common_server_proto.g.dart
   Pub.run('build_runner', arguments: ['build', '--delete-conflicting-outputs']);
 }
+
+class RunWithLoggingException implements Exception {
```
domesticmouse: Hey @devoncarew, any chance of getting something akin to this function merged into `package:grinder`? I'm happy to author PRs if you are happy to review them and re-publish the package...

doc/generated/dartservices.json
```diff
@@ -321,6 +321,18 @@
     "servicesVersion": {
      "type": "string",
      "description": "The dart-services backend version."
+    },
+    "flutterVersion": {
+     "type": "string",
+     "description": "The Flutter SDK version."
+    },
+    "flutterDartVersion": {
```
devoncarew: Will this dart sdk version (the Flutter one) ever be out of sync with the `sdkVersion`? Same with `flutterDartVersionFull` and `sdkVersionFull`?

It'd be confusing to clients about what to show if we return n different versions they could select from. Perhaps just ensure that the versions stay in sync? Fail the CI if we do something where the versions would be out of sync?

tool/grind.dart
devoncarew: Sure, if it fits in with the current API. Some of the utilities in grinder were added before common packages existed (or, there was costs associated with using them on some platforms).

doc/generated/dartservices.json
domesticmouse: Yes, the two dart versions are different, because Flutter for web is not in stable. We are on the `dev` branch, which means we use a pre-release cut of Dart for compiling the Flutter apps. Here's where we are right now:

```json
{
 "sdkVersion": "2.7.1",
 "sdkVersionFull": "2.7.1",
 "runtimeVersion": "2.7.1 (Unknown timestamp) on \"linux_x64\"",
 "appEngineVersion": "2.7.1 (Unknown timestamp) on \"linux_x64\"",
 "servicesVersion": "M1.2",
 "flutterVersion": "1.15.22",
 "flutterDartVersion": "2.8.0",
 "flutterDartVersionFull": "2.8.0-edge.7079c49b054e211b6efac6964574e65635e5da43"
}
```

domesticmouse:
@RedBrogdon I've tested this PR in production, and everything appears happy in DartPad land. Landing as is.

PR #488, opened 2020-03-19
Drop package rpc

domesticmouse:
This depends on https://github.com/dart-lang/dart-pad/pull/1501 landing and being published.


lib/services_dev.dart
```diff
@@ -79,94 +63,33 @@ class EndpointsServer {
...
-    return apiServer
-        .handleHttpApiRequest(apiRequest)
-        .then((HttpApiResponse apiResponse) {
-      // TODO(jcollins-g): use sendApiResponse helper?
-      return Response(apiResponse.status,
-          body: apiResponse.body,
-          headers: Map<String, String>.from(apiResponse.headers));
-    });
+    handler = pipeline.addHandler(
+      (request) => commonServerApi.router.handler(request),
```
johnpryan: I think this could be simplified to `handler = pipeline.addHandler(commonServerApi.router.handler);`

domesticmouse: Good catch, thanks!

domesticmouse:
Status note: this PR is in a holding pattern until we can migrate all the clients across to v2 of the API endpoint, as this PR effectively makes v1 an alias of the v2 endpoint.

domesticmouse:
@RedBrogdon is this now safe to land?

pubspec.lock
```diff
@@ -2,7 +2,7 @@
 # See https://dart.dev/tools/pub/glossary#lockfile
 packages:
   _discoveryapis_commons:
-    dependency: "direct main"
+    dependency: transitive
```
RedBrogdon, line 5: Looks like the appengine package is causing this to hang around.

lib/src/common_server_api.dart
```diff
@@ -0,0 +1,176 @@
...
+
+  @Route.post('$PROTO_API_URL_PREFIX/document')
+  Future<Response> document(Request request, String apiVersion) =>
+      _processRequest(request,
+          decodeFromJSON: (json) =>
+              proto.SourceRequest.create()..mergeFromProto3Json(json),
+          decodeFromProto: (bytes) => proto.SourceRequest.fromBuffer(bytes),
+          transform: _impl.document);
+
+  @Route.post('$PROTO_API_URL_PREFIX/version')
```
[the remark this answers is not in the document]
domesticmouse: Consistency? I think it is used from somewhere, I can't remember where tho.

pubspec.lock
domesticmouse: It doesn't look like it is a direct dependency. Is there a way of finding out the dependency path? 

https://github.com/dart-lang/appengine/blob/master/pubspec.yaml

lib/src/common_server_api.dart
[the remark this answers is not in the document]
domesticmouse: Raised a tracking bug against myself to figure out if I can drop this. https://github.com/dart-lang/dart-pad/issues/1539

PR #490, opened 2020-03-20
Windows fixes, part 1

domesticmouse:
**DO NOT SUBMIT:** Pending product decision regarding https://github.com/dart-lang/dart-pad/issues/1493

This is the start of a fix for https://github.com/dart-lang/dart-pad/issues/1493

This PR is required, but not sufficient, to make `dart-services` run on MS Windows. The rest of the fixes are pending #488 landing.

PR #491, opened 2020-03-21
Windows fixes, part 2

domesticmouse:
**DO NOT SUBMIT:** Pending product decision regarding https://github.com/dart-lang/dart-pad/issues/1493

This PR assumes both #488 and #490 have already landed.


domesticmouse:
We decided not to support MS Windows for deployment of the server at this point in time.

PR #492, opened 2020-03-22
A bunch of minor fixes

domesticmouse:
- Tidy up imports
 - Correct tooling installation docs
 - Fix null handling error in shelf_cors
 - Fix logging in grind script
 - Improve logging around dartfmt errors in travis script

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -12,6 +12,8 @@ analyzer:
 
 linter:
   rules:
+    - directives_ordering
     - prefer_final_fields
     - prefer_final_in_for_each
     - prefer_final_locals
+    - prefer_relative_imports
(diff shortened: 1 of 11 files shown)


analysis_options.yaml
```diff
@@ -12,6 +12,8 @@ analyzer:
 
 linter:
   rules:
+    - directives_ordering
```
RedBrogdon: We should use this one in flutter/samples as well.

domesticmouse: Oooh we have some maintenance to do in the samples prior to the next Flutter stable release.

domesticmouse: And https://github.com/flutter/samples/pull/396

PR #494, opened 2020-03-25
Fix CORS headers for the API

domesticmouse:
fixes: https://github.com/dart-lang/dart-pad/issues/1471


domesticmouse:
PTAL @RedBrogdon

domesticmouse:
PTAL @RedBrogdon 😋

PR #502, opened 2020-04-03
Enable editing dart-services in Gitpod.io

domesticmouse:
Hey @RedBrogdon,

I'm investigating gitpod.io as an easy getting started development environment with zero required downloads. As a starting point, I added the following configuration to `dart-services` to make it possible to get started with developing on it easily. Let me know what you think =)

brett

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -13,7 +13,7 @@ USER gitpod
 
 RUN sudo apt-get update
 RUN sudo apt-get install apt-transport-https
-RUN sudo apt install protobuf-compiler
+RUN sudo apt-get install protobuf-compiler
 RUN sudo apt-get install redis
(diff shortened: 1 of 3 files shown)


RedBrogdon:
Discussed offline with @domesticmouse. The only thing that worries me here is that we'd be taking on additional maintenance by introducing the GitPod integration as an additional requirement.

If we're doing this as a "Brett's life is improved by being able to quickly test things on GitPod, while the rest of us are allowed to completely ignore anything GitPod-related (including breaking the integration)," I'm fine with it.

domesticmouse:
The only real additional maintenance work I can see is keeping the Gitpod Docker file, the travis configuration, and the GAE Flex Docker files in sync. The gitpod configuration can easily be deleted if it bitrots with no downside.

domesticmouse:
@googlebot: rescan

PR #509, opened 2020-04-22
Cached compile response fix

domesticmouse:
TBR: @RedBrogdon


RedBrogdon:
What was the issue here? Was something crashing/failing to parse in dart-pad, or was dart-services failing to send source maps it already had?

domesticmouse:
> What was the issue here? Was something crashing/failing to parse in dart-pad, or was dart-services failing to send source maps it already had?

The issue is in this (simplified) line of code:

```
proto.CompileResponse().sourceMap = returnSourceMap ? resultObj['sourceMap'] as String : null;
```

The underlying Proto generated Dart class throws an exception if you set an attribute to null. The reason I missed this in development was that I was developing without Redis, so I only debugged the straight through code path (which already contained this fix) and not the returned from cache path.

Soon, this will all be caught by the type system. =)

PR #525, opened 2020-05-04
(draft) Add dev version

johnpryan:
re-run Travis to see why the build is hanging


domesticmouse:
This usually passes as a PR then fails after it is submitted to master

PR #536, opened 2020-05-09
Enable https in warmup

domesticmouse:
Turns out Cloud Run doesn't like http traffic (it 302's it across to https) unlike AppEngine Flex. Easy fix is to enable the warmup script to take both http and https urls to warmup.


domesticmouse:
TBR: @johnpryan

PR #547, opened 2020-05-11
Cloud Run fixup

domesticmouse:


lib/services_cloud_run.dart
```diff
@@ -23,19 +23,23 @@ import 'src/shelf_cors.dart' as shelf_cors;
...
   parser.addOption('server-url', defaultsTo: 'http://localhost');
   parser.addOption('redis-url');
-
   final result = parser.parse(args);
-  final port = int.tryParse(result['port'] as String);
+
+  // Cloud Run supplies the port to bind to in the environment.
+  // Allow command line arg to override environment.
+  final port = int.tryParse(result['port'] as String ?? '') ??
+      int.tryParse(Platform.environment['PORT'] ?? '');
```
johnpryan, line 36: I think this should be:

```
  final port = int.tryParse(result['port'] as String ?? null) ??
      int.tryParse(Platform.environment['PORT'] ?? null) ?? '';
```

Otherwise I don't see when the second `int.tryParse` would happen, since the the `int.tryParse(result['port'] as String ?? '')` expression always never returns null

domesticmouse: The reason for this patch was `int.tryParse(null)` throws an exception, so your suggested fix wouldn't prevent the exception. `int.tryParse('')` returns null, however.

lib/services_cloud_run.dart
```diff
@@ -45,10 +49,18 @@ void main(List<String> args) {
...
-  EndpointsServer.serve(sdk, port, redisServerUri)
-      .then((EndpointsServer server) {
-    _logger.info('Listening on port ${server.port}');
-  });
+  _logger.info('''Initializing dart-services:
+    port: $port
+    sdkPath: $sdkPath
+    redisServerUri: $redisServerUri
+    Cloud Run Environment vars:
+    K_SERVICE: ${Platform.environment['K_SERVICE']}
```
kevmoo, line 57: DBC: worth just enumerating environment keys that start with `K_` and print those? It'd be a bit more future-proof

domesticmouse: done.

lib/services_cloud_run.dart
johnpryan: Ah, right.

lib/services_cloud_run.dart
```diff
@@ -45,10 +49,20 @@ void main(List<String> args) {
     if (record.stackTrace != null) print(record.stackTrace);
   });
 
-  EndpointsServer.serve(sdk, port, redisServerUri)
-      .then((EndpointsServer server) {
-    _logger.info('Listening on port ${server.port}');
-  });
+  final cloudRunEnvVars = Platform.environment.entries
+      .where((entry) => entry.key.startsWith('K_'))
```
kevmoo, line 53: @domesticmouse realized there may be an issue here.

Not sure if there is ever any secrets in a `K_` entry that you WOULDN'T want logged.

Something to think about...and sorry for not thinking of it sooner!

domesticmouse: Currently the environment only contains three `K_` keys, as specified in https://cloud.google.com/run/docs/reference/container-contract

I don't expect they will expose secrets through the environment, as we have things like https://cloud.google.com/secret-manager instead

PR #549, opened 2020-05-11
Expands unhealth check to kill instances that fail to restart quickly

RedBrogdon:
Removes CommonServer's _running flag and replaces it with a DateTime called _restartingSince that contains:

* Null if the server is not restarting
* The time the server decided to restart (if the server *is* restarting)

Also adds readonly boolean properties isRestarting and isHealthy to CommonServer based on _restartingSince. The isHealthy flag is true if the server has been trying and failing to restart for more than a half hour.

The GAE readiness and healthchecks are updated to consume the new flags.

### DIFF
--- a/lib/src/common_server_impl.dart
+++ b/lib/src/common_server_impl.dart
@@ -27,6 +27,7 @@ final Logger log = Logger('common_server');
 
 class BadRequest implements Exception {
   String cause;
+
   BadRequest(this.cause);
 }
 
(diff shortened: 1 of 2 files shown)


lib/src/common_server_impl.dart
```diff
@@ -48,8 +49,16 @@ class CommonServerImpl {
       analysisServer.analysisServer != null &&
       flutterAnalysisServer.analysisServer != null;
 
-  bool _running = false;
-  bool get running => _running;
+  // If non-null, this value indicates that server is restarting and holds the
```
johnpryan, line 52: nit: restarting or starting

PR #553, opened 2020-05-12
Add local configuration steps

allentv:
Before unit tests can run, flutter has to be configured. This PR adds the commands to be executed that will configure the Flutter version bundled in the repo.


domesticmouse:
Thanks for the PR!

PR #556, opened 2020-05-13
Adding source to error logs

domesticmouse:


lib/src/common_server_impl.dart
```diff
@@ -331,12 +334,12 @@ class CommonServerImpl {
...
       }
-    }).catchError((dynamic e, dynamic st) {
+    } catch (e, st) {
       if (e is! BadRequest) {
-        log.severe('Error during compile (DDC): $e\n$st');
+        log.severe('Error during compile (DDC) on "$source"', e, st);
       }
-      throw e;
-    });
+      rethrow;
```
RedBrogdon, line 341: This will catch some things the previous `catchError` wasn't wrapping, but since you're rethrowing the exception, I'm guessing that's not an issue.

domesticmouse: We can retune the logging once we see what we generate.

PR #604, opened 2020-09-24
remove the use of the analysis server --dartpad command line option

devoncarew:
- remove the use of the analysis server `--dartpad` command line option

This is a fairly dartpad specific feature of the analysis server - I think we're better off guarding against file paths in the dartpad backend, and not having much if any dartpad specific knowledge in the analysis server.

### DIFF
--- a/lib/src/analysis_server.dart
+++ b/lib/src/analysis_server.dart
@@ -35,6 +35,7 @@ const Duration _ANALYSIS_SERVER_TIMEOUT = Duration(seconds: 35);
 
 class DartAnalysisServerWrapper extends AnalysisServerWrapper {
   Directory _tempProject;
+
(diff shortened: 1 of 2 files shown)


domesticmouse (approved):

devoncarew:
@domesticmouse @johnpryan @RedBrogdon - we're about to remove the `--dartpad` flag from the analysis server; I just want to confirm that this PR is in all the active branches of dartpad? I know it's in master, but I'm not sure if the nnbd branch is active or has this change.

johnpryan:
I'm upgrading the `nnbd` branch with master in the `nnbd2` branch. I deployed in on Friday but compilation broke 

Analysis is working as expected on that branch, so I think it's okay to remove this flag.

devoncarew:
As long as that branch has this commit (2dcc621) we should be good.

PR #605, opened 2020-09-24
Cloud run dark launch

domesticmouse:


lib/src/common_server_impl.dart
```diff
@@ -32,40 +33,170 @@ abstract class ServerContainer {
...
+
+  @override
+  set _analysisServers(AnalysisServersWrapper __analysisServers) =>
+      _wrapped._analysisServers = __analysisServers;
+
+  @override
+  set _compiler(Compiler __compiler) => _wrapped._compiler = __compiler;
+}
+
+final JsonEncoder _jsonEncoder = const JsonEncoder.withIndent(' ');
```
johnpryan, line 164: why is this using one space for indentation? I usually see a two or four, or a tab character

johnpryan (approved):
LGTM

lib/src/common_server_impl.dart
domesticmouse: Updated to use 2 characters. Not that it really matters, it's used for the Cloud Run request. No human will ever see it =)

PR #618, opened 2020-11-16
Drop beta channel from CI

domesticmouse:
Dart `beta` is now 2.12, and this requires changing tests for Dart Null Safety, per #614


domesticmouse:
TBR: @johnpryan 

This is a breakfix for the broken CI. =)

johnpryan:
LGTM

PR #620, opened 2020-11-20
Adding the Flutter Sunflower test

domesticmouse:


domesticmouse:
Hey @devoncarew, I have a need to run the CompileDDC in non null safe mode. Is there command line arguments I can hand `dartdevc` to do the equivalent of setting `sdk: ^2.10.4` in a `pubspec.yaml`?

Here's the context: https://github.com/domesticmouse/dart-services/blob/flutter-sunflower-test/lib/src/compiler.dart#L140

domesticmouse:
Pulling failure test log:

```
01:03 +75: compiler compileDDC with Flutter Sunflower

[/tmp/dartpadHCRFPT/main.dart:50:30: Error: The argument type 'num' can't be assigned to the parameter type 'double'.
    canvas.drawCircle(Offset(x, y), seedRadius, paint);
                             ^
/tmp/dartpadHCRFPT/main.dart:50:33: Error: The argument type 'num' can't be assigned to the parameter type 'double'.
    canvas.drawCircle(Offset(x, y), seedRadius, paint);
                                ^
]

01:03 +75 -1: compiler compileDDC with Flutter Sunflower [E]

  Expected: <true>
    Actual: <false>
  

  package:test_api/src/frontend/expect.dart 154:30               fail
  package:test_api/src/frontend/expect.dart 148:3                _expect
  package:test_api/src/frontend/expect.dart 60:3                 expect
  test/compiler_test.dart 72:7                                   defineTests.<fn>.<fn>
```

devoncarew:
> Hey @devoncarew, I have a need to run the CompileDDC in non null safe mode. Is there command line arguments I can hand `dartdevc` to do the equivalent of setting `sdk: ^2.10.4` in a `pubspec.yaml`?

Probably? You might also try the pubspec route. cc'ing @nshahan who would know.

domesticmouse:
> You might also try the pubspec route.

I tried populating the directory with a `pubspec.yaml`, but it looked like the bazel'd dartdevc process ignored it. If there is a way of getting pubspec uptake then that would WFM.

devoncarew:
> I tried populating the directory with a `pubspec.yaml`, but it looked like the bazel'd dartdevc process ignored it.

I think that's unexpected.

The pubspec.yaml file has an sdk constraint that opts it out of nnbd? Thats < 2.12.0 (like, `>=2.10.0 <3.0.0`).

If you run pub on that, it'll write the correct language version into the .dart_tool/package_config.json file. If we don't run pub, we'd just need to create a package_config.json file by hand (its a simple file format).

devoncarew:
> The argument type 'num' can't be assigned to the parameter type 'double'.

I'm wondering if its not null safety related however. I'd try playing with the types there - that looks downcast related.

RedBrogdon:
I don't believe this is directly null safety-related. I want to say there was a change in SDK behavior relating to this recently, though -- it used to automatically cast from num to double for calls like that and now it won't.

I had to update some of the other gists, IIRC, and figured it was better to force the explicit cast anyway.

domesticmouse:
I've added the code for the other Flutter samples from the drop down, and here's their failure traces:

### Flutter Counter sample

```
Output for compiler compileDDC with Flutter Counter
[/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadgy24wZ/main.dart:20:19: Error: Optional parameter 'key' should have a default value because its type 'Key' doesn't allow null.
 - 'Key' is from 'package:flutter/src/foundation/key.dart' ('flutter/packages/flutter/lib/src/foundation/key.dart').
  MyHomePage({Key key, this.title}) : super(key: key);
                  ^^^
/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadgy24wZ/main.dart:20:29: Error: Optional parameter 'title' should have a default value because its type 'String' doesn't allow null.
  MyHomePage({Key key, this.title}) : super(key: key);
                            ^^^^^
]
Expected: <true>
  Actual: <false>

package:test_api              expect
test/compiler_test.dart 72:7  defineTests.<fn>.<fn>
```

### Draggable Card sample

```
Output for compiler compileDDC with Flutter Draggable Card
[/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadlVX4X9/main.dart:31:23: Error: Optional parameter 'child' should have a default value because its type 'Widget' doesn't allow null.
 - 'Widget' is from 'package:flutter/src/widgets/framework.dart' ('flutter/packages/flutter/lib/src/widgets/framework.dart').
  DraggableCard({this.child});
                      ^^^^^
/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadlVX4X9/main.dart:39:23: Error: Field '_controller' should be initialized because its type 'AnimationController' doesn't allow null.
 - 'AnimationController' is from 'package:flutter/src/animation/animation_controller.dart' ('flutter/packages/flutter/lib/src/animation/animation_controller.dart').
  AnimationController _controller;
                      ^^^^^^^^^^^
/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadlVX4X9/main.dart:41:24: Error: Field '_animation' should be initialized because its type 'Animation<Alignment>' doesn't allow null.
 - 'Animation' is from 'package:flutter/src/animation/animation.dart' ('flutter/packages/flutter/lib/src/animation/animation.dart').
 - 'Alignment' is from 'package:flutter/src/painting/alignment.dart' ('flutter/packages/flutter/lib/src/painting/alignment.dart').
  Animation<Alignment> _animation;
                       ^^^^^^^^^^
]
Expected: <true>
  Actual: <false>

package:test_api              expect
test/compiler_test.dart 94:7  defineTests.<fn>.<fn>
```

### Implicit Animations sample

```
Output for compiler compileDDC with Flutter Implicit Animations
[/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadNo1ZcU/main.dart:7:10: Error: Field 'size' should be initialized because its type 'double' doesn't allow null.
  double size;
         ^^^^
/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadNo1ZcU/main.dart:8:9: Error: Field 'color' should be initialized because its type 'Color' doesn't allow null.
 - 'Color' is from 'dart:ui'.
  Color color;
        ^^^^^
/var/folders/yx/p5zw_q4x561g718ttqx7bs4c003nrc/T/dartpadNo1ZcU/main.dart:9:13: Error: Field 'alignment' should be initialized because its type 'Alignment' doesn't allow null.
 - 'Alignment' is from 'package:flutter/src/painting/alignment.dart' ('flutter/packages/flutter/lib/src/painting/alignment.dart').
  Alignment alignment;
            ^^^^^^^^^
]
Expected: <true>
  Actual: <false>

package:test_api               expect
test/compiler_test.dart 106:7  defineTests.<fn>.<fn>
```

nshahan:
> Hey @devoncarew, I have a need to run the CompileDDC in non null safe mode. Is there command line arguments I can hand dartdevc to do the equivalent of setting sdk: ^2.10.4 in a pubspec.yaml?

There isn't any command line flag for bin/dartdevc to override the language version detection. In typical builds the language version is determined by the CFE through detection from the pubspec/package.json or language version comments at the beginning of the file.

Out of curiosity are you running bin/dartdevc directly or some other tool that runs DDC? What is the command and all the arguments you are using?

> I don't believe this is directly null safety-related. I want to say there was a change in SDK behavior relating to this recently, though -- it used to automatically cast from num to double for calls like that and now it won't.

I agree this is a language version issue, implicit downcasts (from num to double in this case) have been removed from the language. It's not part of the null safety feature per se (although arguably it is somewhat required to avoid getting nulls where you don't want them) but the change is included in the null safety release. Setting the correct language version for these files is the fix you would need if you have not migrated the files yet. During the migration making the casts explicit might be the right choice since the casts were already there implicitly at runtime.

domesticmouse:
Hey @nshahan,

We are running `dartdevc` inside of a bazel worker. The command line args from a test instance on my laptop are something like this:

```
About to exec "/Users/brettmorgan/Documents/GitHub/dart-services/flutter/bin/cache/dart-sdk/bin/dartdevc --modules=amd -o /var/folders/6r/rk5z2g4521gfhwf4jw704pvh003nrc/T/dartpadhWBhdw/main.dart.js --module-name dartpad_main /var/folders/6r/rk5z2g4521gfhwf4jw704pvh003nrc/T/dartpadhWBhdw/bootstrap.dart --packages=/var/folders/6r/rk5z2g4521gfhwf4jw704pvh003nrc/T/dartpadW4uPC1/.packages"
```

Hey @devoncarew, 

We write out a `pubspec.yaml` during spin up of the `FlutterWebManager`, and run `flutter pub get` in that directory. The generated pubspec has the following environment constraint:

```
environment:
  sdk: '>=2.10.0 <3.0.0'
```

Hey @RedBrogdon,

I can't see a way to make this release run the Flutter tooled `compileDDC` in a non null safe way. As a work around, we can hold off updating the server binaries until we are ready to have everything non null.

domesticmouse:
TBR: @RedBrogdon 

With #621 landed, I'm going to land this PR to mainline the test suite. Then I'll open a new PR to investigate re-landing Flutter 1.24.

PR #619, opened 2020-11-18
Flutter 1.24

domesticmouse:
This PR has some work arounds for the Dart 2.12 beta bundled into Flutter 1.24 beta.

PR #621, opened 2020-11-23
Revert "Flutter 1.24 (#619)"

domesticmouse:
Revert to Flutter 1.23, without dropping the updated tests


domesticmouse:
TBR @johnpryan, this is effectively a breakfix.

PR #625, opened 2020-12-17
Add Dart dev and beta channels

domesticmouse:


domesticmouse:
TBR: @johnpryan

.github/workflows/dart.yml
```diff
@@ -6,15 +6,25 @@ on:
...
   # against Dart stable and beta.
   analyze:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
-        sdk: [stable]
+        include:
+          - sdk: dev
+            allow_failure: true
```
johnpryan, line 23: I think we should *not* allow failures, and instead remove these checks from the branch protection rule. See my comment here: https://github.com/flutter/samples/pull/629#issuecomment-747606716

PR #626, opened 2020-12-17
Drop Travis

domesticmouse:


domesticmouse:
Hey @RedBrogdon, it looks like the travis tests are required on this repo, which is a problem from the point of view of removing them. Any ideas on where to configure which checks are required?

parlough (approved):

RedBrogdon:
In the repo settings under "branches" you can select which presubmit checks you consider mandatory. If you don't have access to the Settings, we should get you added to whatever org is needed.

In the meantime, I removed travis from the list and replaced it with the stable version of the GH Actions. Let me know if you'd prefer a different configuration.

domesticmouse:
This configuration LGTM. And yeah, I don't have access to configure presubmit checks under settings -> branches. =)

PR #628, opened 2020-12-17
Drop continue_on_failure

domesticmouse:


domesticmouse:
Hey @RedBrogdon, looks like we will need to update the required checks to land this change =)

johnpryan (approved):
This looks good to me, as long as everyone likes the approach. (This is my preference but it does require tweaking the required checks)

PR #630, opened 2020-12-29
misc changes to dart-services

devoncarew:
misc changes to dart-services:
- update text in comments
- some updates to the analysis_options file
- reduce the number of deps in the pubspec file


analysis_options.yaml
```diff
@@ -4,11 +4,11 @@ analyzer:
...
-    - "dart-sdk/**"
-    - "third_party/**"
-    - "doc/generated/**"
-    - "flutter/**"
-    - "lib/src/protos/**"
+    - dart-sdk/**
+    - doc/generated/**
+    - flutter/**
+    # TODO: This seems to be hiding ~2 dozen legitimate analysis issues.
+    - lib/src/protos/**
```
domesticmouse, line 11: The issues here are in generated files. Feel free to assign me a bug to get appropriate changes made to the protobuf generator templates

pubspec.yaml
```diff
@@ -33,10 +30,8 @@ dev_dependencies:
   grinder: ^0.8.0
   mock_request: ^1.0.7
   pedantic: ^1.9.0
-  shelf_router_generator: ^0.7.0+1
```
domesticmouse, line 36: I'm surprised this is no longer required, I thought I was using it to generate shelf routes.

domesticmouse (approved):

pubspec.yaml
devoncarew: I don't see any references to it in the repo (via `git grep shelf_router_generator`).

analysis_options.yaml
devoncarew: Sure! https://github.com/dart-lang/dart-pad/issues/1712

pubspec.yaml
domesticmouse: I believe it is required to generated `lib/src/common_server_api.g.dart`, which probably shouldn't be checked in to the repo.

devoncarew: Hmm; perhaps instead of a dev. dependency, this could be something which people generating the source need to pub global activate?

domesticmouse: reverting in #633

PR #636, opened 2021-01-07
Don't allow dart:io

domesticmouse:
fixes: https://github.com/dart-lang/dart-pad/issues/1692
fixes: https://github.com/dart-lang/dart-pad/issues/1615


domesticmouse:
TBR: @RedBrogdon

PR #638, opened 2021-01-13
minor lint related changes

devoncarew:
- minor lint related changes

These lint related changes are the delta from trying out the proposed new dart rule set.

cc @pq

### DIFF
--- a/lib/services_gae.dart
+++ b/lib/services_gae.dart
@@ -154,7 +154,7 @@ class GaeServer {
       try {
         final tempDir = await io.Directory.systemTemp.createTemp('healthz');
         try {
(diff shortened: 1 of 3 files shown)


lib/services_gae.dart
```diff
@@ -154,7 +154,7 @@ class GaeServer {
       try {
         final tempDir = await io.Directory.systemTemp.createTemp('healthz');
         try {
-          final file = await io.File('${tempDir.path}/livecheck.txt');
+          final file = io.File('${tempDir.path}/livecheck.txt');
```
pq, line 157: nice catch!

pq (approved):
👍

PR #642, opened 2021-01-21
Mark other VM imports as unsupported

parlough:
None of these have runtime support on web platforms, similar to `dart:io`.

### DIFF
--- a/lib/src/flutter_web.dart
+++ b/lib/src/flutter_web.dart
@@ -24,9 +24,16 @@ class FlutterWebManager {
     return path.join('artifacts', 'flutter_web.dill');
   }
 
-  static final Set<String> _flutterWebImportPrefixes = <String>{
+  static const Set<String> _flutterWebImportPrefixes = {
     'package:flutter',
     'dart:ui',
+  };
+
+  static const Set<String> _dartVmImports = {
+    'dart:ffi',
+    'dart:io',
+    'dart:isolate',
+    'dart:mirrors',
   };
 
   bool usesFlutterWeb(Set<String> imports) {
@@ -43,8 +50,8 @@ class FlutterWebManager {
 
   String getUnsupportedImport(Set<String> imports) {
     for (final import in imports) {
-      // All dart: imports are ok, except dart:io
-      if (import.startsWith('dart:') && import != 'dart:io') {
+      // All non-VM dart: imports are ok
+      if (import.startsWith('dart:') && !_dartVmImports.contains(import)) {
         continue;
       }
 
--- a/lib/src/flutter_web.dart
+++ b/lib/src/flutter_web.dart
@@ -29,11 +29,22 @@ class FlutterWebManager {
     'dart:ui',
   };
 
-  static const Set<String> _dartVmImports = {
-    'dart:ffi',
-    'dart:io',
-    'dart:isolate',
-    'dart:mirrors',
+  static const Set<String> _allowedDartImports = {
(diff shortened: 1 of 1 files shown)


domesticmouse:
@kevmoo does this look about right?

kevmoo (approved):

lib/src/flutter_web.dart
```diff
@@ -43,8 +50,8 @@ class FlutterWebManager {
 
   String getUnsupportedImport(Set<String> imports) {
     for (final import in imports) {
-      // All dart: imports are ok, except dart:io
-      if (import.startsWith('dart:') && import != 'dart:io') {
+      // All non-VM dart: imports are ok
+      if (import.startsWith('dart:') && !_dartVmImports.contains(import)) {
```
kevmoo, line 54: although it might be better to do an allow list

kevmoo: image

domesticmouse: @parlough I concur with @kevmoo's suggestion to use a whitelist.

@kevmoo is there a source for the list of supported libraries? It'd be useful to link to this location in the source code.

kevmoo: dartdoc pulls it from somewhere - https://api.dart.dev/stable/2.10.5/index.html – @jcollins-g ?

kevmoo: *cough* allowlist *cough* 😄

parlough: If we're going with the `allowList` don't we need to include the web libraries as well here? For the HTML version of DartPad at least but even the flutter web version can use them, but a separate `info` is provided suggesting to only use them in packages.

I've gone ahead and included them, but let me know if you'd like to see any changes.

I didn't know how to programmatically retrieve the web-supported libraries, perhaps the analyzer can do so. As for the `dartdoc` functionality that seems to be due to Dartdoc categories, as can be seen here - https://github.com/dart-lang/sdk/blob/master/sdk/lib/mirrors/mirrors.dart#L56

domesticmouse: Good point, we need to add `dart:html` and friends to the allow list, or a bunch of our samples will die an ignoble death. I believe we should add the following:

- `dart:html`
- `dart:indexed_db`
- `dart:js`
- `dart:js_util`
- `dart:svg`
- `dart:web_gl`
- `dart:web_sql`

Adding @johnpryan as sanity reviewer =)

lib/src/flutter_web.dart
```diff
@@ -24,11 +24,32 @@ class FlutterWebManager {
...
+    'dart:typed_data',
+    'dart:html',
+    'dart:indexed_db',
+    'dart:js',
+    'dart:js_util',
+    'dart:svg',
+    'dart:web_audio',
+    'dart:web_gl',
+    'dart:web_sql',
+    'dart:ui',
```
domesticmouse, line 50: Add dart web imports plz

domesticmouse (changes requested):

lib/src/flutter_web.dart
parlough: I believe they're in the list already, am I missing any?

domesticmouse: Whups, my bad. LGTM

domesticmouse (approved):

domesticmouse:
TBR @johnpryan

PR #643, opened 2021-01-21
Remove old TODO and set default parameter value

parlough:
This `TODO` seemed to already be addressed and I think the current name works well.

I also updated the optional named parameter to assign the value in the declaration, which will be necessary with the eventually transition to null safety anyway. The `dartfmt` looks a bit weird in this case, but oh well..


domesticmouse (approved):
Thanks!

PR #649, opened 2021-01-27
Enforce two strict analysis modes

srawlins:
I like these changes for a few specific reasons:

* `Future` => `Future<void>` ensures with static analysis you won't try to use the result.
* `Future` => `Future<AnalysisServer>` changes any method/field access on the result from dynamic dispatch to static dispatch; safer, smaller, faster.
* `Future` => `Future<dynamic>` is not really a functional improvement; it lucks ugly. But it takes code that was already using dynamic (just secretly) and makes it explicit, so that developers are aware of dynamicism in their code.

Some changes like `{}` => `<dynamic, dynamic>{}` are pretty ugly. I hope that with generalized typedefs, some styles come out like `typedef JsonMap = Map<dynamic, dynamic>`, etc.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,6 +2,9 @@
 # include: package:pedantic/analysis_options.1.9.0.yaml
 
 analyzer:
+  language:
+    strict-inference: true
+    strict-raw-types: true
   strong-mode:
     implicit-casts: false
   exclude:
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -3,7 +3,6 @@
 
 analyzer:
   language:
-    strict-inference: true
     strict-raw-types: true
   strong-mode:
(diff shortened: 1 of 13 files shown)


domesticmouse (approved):

parlough (approved):

domesticmouse:
This change is impacting the compilation phase. That's probably bad. This may need to wait until after we implement the single sdk change.

/cc @devoncarew

devoncarew:
DBC: do we want to update the flutter submodule as part of this change? It may be best to keep that separate from this PR.

srawlins:
I think they should definitely be different changes.

srawlins:
OK it was the analysis_options causing problems. This is ready for review now.

domesticmouse (approved):

PR #652, opened 2021-02-02
add the ability to provision a flutter sdk via a configuration file

devoncarew:
- add the ability to provision a flutter sdk via a configuration file
- general work towards https://github.com/dart-lang/dart-pad/issues/1715

This uses `//flutter-sdk-version.yaml` as the configuration file and writes the sdk to `//flutter-sdk/`. There's a manual option to update the local sdk (running `tool/update_sdk.dart`). This SDK isn't hooked into dart-services - the server - at all, and doesn't take the place of the current flutter sdk at `//flutter/` (or the dart sdk from the docker image). I think we'll want to move over to using this newer sdk setup in stages - make sure that our existing deploys and in-progress version upgrades aren't negatively impacted.

### DIFF
--- a/lib/src/sdk_manager.dart
+++ b/lib/src/sdk_manager.dart
@@ -3,9 +3,12 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:convert' show utf8;
 import 'dart:io';
 
+import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
+import 'package:yaml/yaml.dart';
 
 /// Generally, this should be a singleton instance (it's a heavy-weight object).
 class SdkManager {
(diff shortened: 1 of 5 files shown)


lib/src/sdk_manager.dart
```diff
@@ -94,3 +93,215 @@ class FlutterSdk extends Sdk {
...
+
+    return sdk;
+  }
+}
+
+class DownloadedFlutterSdk extends Sdk {
+  @override
+  Future<void> init() async {
+    // flutter --version takes ~28s
+    await _execLog('bin/flutter', ['--version'], flutterSdkPath);
```
domesticmouse, line 210: Can we pull this out of the serving binary and do it in the grind script instead? Having the server come up without being able to serve is bad.

Alternatively, delaying opening the port on 8080 until after flutter is ready is acceptable for Cloud Run.

domesticmouse (approved):

lib/src/sdk_manager.dart
devoncarew: This is part of the sdk init; the other sdk subclasses have this as well (though this one is potentially much longer running).

This will only take time if the sdk has not yet been initialized; if it has been this will just run in the time it takes to execute `flutter --version`.

We should ensure the server doesn't report as ready to serve until everything's been inited (getting the flutter sdk, the dart sdk, and the engine artifacts). I assume that's what the grind script does? In any case, I agree that we want to do all the expensive bits before we report the server as ready, but that's really the responsibility of the code that uses this sdk class, not of the class itself.

domesticmouse: Sgtm. Once you land this change I'll start migrating across to the new sdk for Flutter compilation.

PR #653, opened 2021-02-04
Drop flutter submodule

domesticmouse:
Devon, I believe this is a fairly complete migration to the downloaded Flutter SDK. I'm going to land this PR as is so I can test the rest of the build and deploy to Cloud Run workflow.

TBR: @devoncarew

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,6 @@
 # Keep aligned with min SDK in pubspec.yaml and .github/workflows/dart.yml.
 FROM google/dart:2.10.4
 
-# The specific commit that dart-services should use. This should be kept
-# in sync with the flutter submodule in the dart-services repo.
-# To retrieve this value, please run the following in your closest shell:
-#
-# $ (cd flutter && git rev-parse HEAD)
-ARG FLUTTER_COMMIT=5d36f2e7f5387b6c751449258ade8e4e6edf99be
-
 # We install unzip and remove the apt-index again to keep the
 # docker image diff small.
 RUN apt-get update && \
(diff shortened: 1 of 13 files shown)


Dockerfile
```diff
@@ -31,18 +24,11 @@ RUN pub get --offline
...
-# Clone the flutter repo and set it to the same commit as the flutter submodule.
-RUN git clone https://github.com/flutter/flutter.git
-RUN cd flutter && git checkout $FLUTTER_COMMIT
-
 # Set the Flutter SDK up for web compilation.
-RUN flutter/bin/flutter doctor
-RUN flutter/bin/flutter config --enable-web
-RUN flutter/bin/flutter precache --web --no-android --no-ios --no-linux \
-  --no-windows --no-macos --no-fuchsia
+RUN dart pub run grinder setup-flutter-sdk
```
devoncarew, line 28: You _should_ be able to shorten this slightly to just `dart run grinder setup-flutter-sdk` (using the new dart tool).

flutter-sdk-version.yaml
```diff
@@ -11,5 +11,5 @@
 # In order to update the SDK manually, run 'dart tool/update_sdk.dart'.
 
 flutter_sdk:
-  channel: beta
-  # version: 1.25.0-8.1.pre
+  # channel: beta
+  version: 1.25.0-8.3.pre
```
devoncarew, line 15: Not for this PR, but going forward, do we want master to track a specific flutter version? A flutter channel? Probably worth thinking about for when we want to do something like deploy dev, beta, and stable sdk versions of dartpad.

tool/grind.dart
```diff
@@ -246,33 +244,30 @@ Future _buildStorageArtifacts(Directory dir) async {
 }
 
 @Task('Delete, re-download, and reinitialize the Flutter submodule.')
-void setupFlutterSubmodule() async {
-  final flutterDir = Directory('flutter');
+void setupFlutterSdk() async {
+  await flutterSdkPath.delete(recursive: true);
```
devoncarew, line 248: I don't think it's necessary to nuke the sdk location each time. The git code I wrote should be able to re-target a sdk checkout to the version / channel we want.

flutter-sdk-version.yaml
domesticmouse: I need a commit to push the Cloud Build pipeline to build and deploy to Cloud Run. So, for the time being I'm being explicit with version numbers as a way of making sure I have an explicit build trigger PR.

In the future I'll need a way of having stable, beta, and dev builds, probably on three separate branches?

Dockerfile
domesticmouse: Fixed in https://github.com/dart-lang/dart-services/pull/645. Turns out I had a chicken and egg problem which took some engineering to fix.

tool/grind.dart
domesticmouse: Also fixed in https://github.com/dart-lang/dart-services/pull/645.

PR #654, opened 2021-02-04
Fixup README and Dockerfiles

domesticmouse:
Fixup Dockerfiles

TBR: @devoncarew

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,13 @@ RUN pub get --offline
 ENV PATH="/home/dart/.pub-cache/bin:${PATH}"
 
 # Set the Flutter SDK up for web compilation.
-RUN dart pub run grinder setup-flutter-sdk
+RUN dart run tool/update_sdk.dart
+
+# Set the Flutter SDK up for web compilation.	# Set the Flutter SDK up for web compilation.
+RUN flutter-sdk/bin/flutter doctor
+RUN flutter-sdk/bin/flutter config --enable-web
+RUN flutter-sdk/bin/flutter precache --web --no-android --no-ios --no-linux \
+  --no-windows --no-macos --no-fuchsia
 
 # Build the dill file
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -25,7 +25,13 @@ RUN pub get --offline
 ENV PATH="/home/dart/.pub-cache/bin:${PATH}"
 
 # Set the Flutter SDK up for web compilation.
-RUN dart pub run grinder setup-flutter-sdk
+RUN dart run tool/update_sdk.dart
+
+# Set the Flutter SDK up for web compilation.	# Set the Flutter SDK up for web compilation.
+RUN flutter-sdk/bin/flutter doctor
```
devoncarew, line 31: We may want to run `flutter config --no-analytics` before we run other flutter commands.

README.md
```diff
@@ -36,12 +36,11 @@ $ pub global activate protoc_plugin
...
 ## Initialise Flutter
 
-The Flutter submodule in the git repo needs to also be downloaded and setup.
+The Flutter SDK needs to be downloaded and setup.
 
 ```bash
-$ git submodule update --init
-$ ./flutter/bin/flutter doctor -v
-$ ./flutter/bin/flutter/bin/flutter config --enable-web
+$ dart pub install
```
devoncarew, line 42: what does `pub install` do?

Dockerfile
domesticmouse: These Dockerfiles only get run by the Cloud Build tooling, thus only once per commit. I doubt it's going to change the stats =)

README.md
domesticmouse: A clean checkout has none of the dart deps downloaded, and `grind deploy` fails in that case. Running `dart pub install` pulls these dependencies, and then `grind deploy` runs clean.

devoncarew: Ah, `pub install` is an alias for `pub get` - I wasn't aware of it.

PR #656, opened 2021-02-04
remove some duplicated code

devoncarew:
- remove some duplicated code (the `FlutterSdk` class is the only one that people outside of the sdk_manager.dart file should know about, and the `_DownloadedFlutterSdk` is now private and an internal helper class for that file)

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -14,6 +14,7 @@ analyzer:
 
 linter:
   rules:
+    - await_only_futures
     - directives_ordering
     - prefer_final_fields
     - prefer_final_in_for_each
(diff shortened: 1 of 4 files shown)


domesticmouse (approved):

domesticmouse:
@devoncarew Am I safe to delete `remove_duplicated_branch`?

devoncarew:
> @devoncarew Am I safe to delete `remove_duplicated_branch`?

Sure! I could also move to working off a fork of this repo.

domesticmouse:
I'm not worried either way, I just prefer not to have dead branches =)

PR #657, opened 2021-02-05
Use Flutter SDK for both Flutter and Dart snippets

domesticmouse:
Re-built https://github.com/dart-lang/dart-services/pull/655 after tidyup

fixes: dart-lang/dart-pad#1714

### DIFF
--- a/lib/src/sdk_manager.dart
+++ b/lib/src/sdk_manager.dart
@@ -12,24 +12,14 @@ import 'package:yaml/yaml.dart';
 
 /// Generally, this should be a singleton instance (it's a heavy-weight object).
 class SdkManager {
-  static Sdk get sdk => _sdk ?? (_sdk = PlatformSdk());
+  static FlutterSdk get sdk => _sdk ?? (_sdk = FlutterSdk(flutterSdkPath));
 
-  static FlutterSdk get flutterSdk =>
-      _flutterSdk ?? (_flutterSdk = FlutterSdk(flutterSdkPath));
-
-  static void setSdk(Sdk value) {
+  static void setSdk(FlutterSdk sdk) {
     _sdk = sdk;
   }
 
-  static void setFlutterSdk(FlutterSdk value) {
-    _flutterSdk = value;
-  }
-
-  // The installed Dart SDK.
-  static Sdk _sdk;
-
   // The installed Flutter SDK.
-  static FlutterSdk _flutterSdk;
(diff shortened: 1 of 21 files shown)


lib/src/sdk_manager.dart
```diff
@@ -12,24 +12,14 @@ import 'package:yaml/yaml.dart';
...
-  static void setFlutterSdk(FlutterSdk value) {
-    _flutterSdk = value;
-  }
-
-  // The installed Dart SDK.
-  static Sdk _sdk;
-
   // The installed Flutter SDK.
-  static FlutterSdk _flutterSdk;
+  static FlutterSdk _sdk;
```
johnpryan, line 22: Since there's only one SDK now, is the `SdkManager` class still useful?

lib/src/sdk_manager.dart
```diff
@@ -51,19 +41,6 @@ abstract class Sdk {
...
-  @override
-  Future<void> init() => Future.value();
-
-  @override
-  String get versionFull =>
-      File(path.join(sdkPath, 'version')).readAsStringSync().trim();
-
-  @override
-  String get sdkPath => path.dirname(path.dirname(Platform.resolvedExecutable));
-}
```
johnpryan, line 65: Since there's only one implementation of `Sdk` now, we can probably remove it too.

johnpryan (approved):
looks good, just a few questions

lib/src/sdk_manager.dart
domesticmouse: Want to do me a favour and raise this as a tech debt item on the tracker and assign it to me? I doubt I'm going to have time to re-arrange the deck chairs before March 3rd when this PR is going to land.

lib/src/sdk_manager.dart
domesticmouse: Ditto

lib/src/sdk_manager.dart
johnpryan: 👍

johnpryan: https://github.com/dart-lang/dart-pad/issues/1768

lib/src/sdk_manager.dart
johnpryan: Added this to https://github.com/dart-lang/dart-pad/issues/1768 as well

PR #663, opened 2021-02-19
Add URL, correction, and diagnostic messages to AnalysisIssue messages

johnpryan:
This enables https://github.com/dart-lang/dart-pad/issues/1676

PR #664, opened 2021-02-23
Strip file paths from analyzer messages

johnpryan:
This adds path stripping on the server side so that filesystem information isn't sent over the wire. I added some tests that cover the primary cases I'm aware of, but we might need additional test cases in the future.

This depends on https://github.com/dart-lang/dart-services/pull/663, and needs to be deployed with https://github.com/dart-lang/dart-pad/pull/1770

### DIFF
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -1,0 +1,6 @@
+import 'package:path/path.dart' as path;
+
+String stripFilePaths(String s) {
+  return s.replaceAllMapped(
+      RegExp(r'/\S*'), (match) => path.basename(match.group(0)));
+}
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -1,13 +1,18 @@
 import 'package:path/path.dart' as path;
 
 String stripFilePaths(String s) {
-  return s.replaceAllMapped(RegExp(r'/\S*'), (match) {
+  return s.replaceAllMapped(RegExp(r'(?:package:?)?[a-z]*\/\S*'), (match) {
     final urlString = match.group(0);
     final isDartPath = path.split(urlString).contains('lib');
+    final isPackagePath = urlString.contains('package:');
     final basename = path.basename(urlString);
 
     if (isDartPath) {
       return path.join('dart:core', basename);
+    }
+
+    if (isPackagePath) {
+      return urlString;
     }
     return basename;
   });
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -1,6 +1,14 @@
 import 'package:path/path.dart' as path;
 
 String stripFilePaths(String s) {
-  return s.replaceAllMapped(
-      RegExp(r'/\S*'), (match) => path.basename(match.group(0)));
+  return s.replaceAllMapped(RegExp(r'/\S*'), (match) {
+    final urlString = match.group(0);
+    final isDartPath = path.split(urlString).contains('lib');
+    final basename = path.basename(urlString);
+
+    if (isDartPath) {
+      return path.join('dart:core', basename);
+    }
+    return basename;
+  });
 }
--- a/lib/src/utils.dart
+++ b/lib/src/utils.dart
@@ -3,9 +3,15 @@ import 'package:path/path.dart' as path;
 String stripFilePaths(String s) {
   return s.replaceAllMapped(RegExp(r'(?:package:?)?[a-z]*\/\S*'), (match) {
     final urlString = match.group(0);
-    final isDartPath = path.split(urlString).contains('lib');
+    final pathComponents = path.split(urlString);
+    final isDartPath = pathComponents.contains('lib');
+    final isFlutterPath = isDartPath && pathComponents.contains('flutter');
     final isPackagePath = urlString.contains('package:');
     final basename = path.basename(urlString);
+
+    if (isFlutterPath) {
+      return path.join('package:flutter', basename);
+    }
 
     if (isDartPath) {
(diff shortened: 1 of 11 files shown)


johnpryan:
cc: @devoncarew

domesticmouse (approved):

lib/src/utils.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:path/path.dart' as path;
+
+String stripFilePaths(String s) {
+  return s.replaceAllMapped(RegExp(r'(?:package:?)?[a-z]*\/\S*'), (match) {
```
devoncarew, line 4: Perhaps extract out to a const or final top-level variable?

I often add some line docs around more complicated regexs as examples of things they are intended to match (it's just often hard to read regexs after their authored).

test/utils_test.dart
```diff
@@ -0,0 +1,16 @@
+library utils.tests;
+
+import 'package:dart_services/src/utils.dart';
+import 'package:expected_output/expected_output.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('utils.stripFilePaths', () {
+    for (final dataCase in dataCasesUnder(library: #utils.tests)) {
```
devoncarew, line 9: Interesting, I wasn't familiar with this test library.

lib/src/utils.dart
devoncarew: I'm not sure the code below does what we want. I think we're only concerned about file paths? Not dart: or package: ones? Will the code below convert `lib/foo.dart` to `dart:core/foo.dart`?

We might want to pass dart: and package: based paths back as is, and convert all others from `foo/bar/baz.dart` to `baz.dart`?

lib/src/utils.dart
```diff
@@ -0,0 +1,25 @@
...
+    final urlString = match.group(0);
+    final pathComponents = path.split(urlString);
+    final isDartPath = pathComponents.contains('lib');
```
devoncarew, line 7: Perhaps expand this check to `/lib/core/`? The path we're getting back from the analysis server is unfortunate - that we have to patch it up after the fact to get a `dart:` path.

test/utils_test.unit
```diff
@@ -0,0 +1,20 @@
+>>> Strips temporary directory paths
+List is defined in /var/folders/4p/y54w9nqj0_n6ryqwn7lxqz6800m6cw/T/DartAnalysisWrapperintLAw/main.dart
+<<<
+List is defined in main.dart
+>>> Strips Dart SDK paths
+List is defined in /Users/ryjohn/.dvm/darts/2.10.5/lib/core/list.dart
```
devoncarew, line 6: I think it's a bug that the analysis server returns this path - it should likely instead return `dart:core/list.dart`.

bwilkerson: It was an intentional design choice. The analysis server protocol always uses absolute and normalized file paths in both directions. Clients should not need to convert from URIs to file paths, or vice versa, because the logic behind it is too complex. There is no requirement that clients should display the file path as the link text, nor was there any expectation that they'd want to do so; the message or some standard text would do just as well. However, if we have a client that wants to display some URI-like text as the link text, then we should add a field for this to the protocol.

devoncarew: I think that makes sense for things that are returned as part of API - for file paths returned to a higher level client in a structured result. Note that these are script references embedded in strings that are user facing; the user will understand `dart:core/list.dart` much better than something like `/Users/ryjohn/.dvm/darts/2.10.5/lib/core/list.dart`. That leaks a lot of info about where the sdk is, the internal paths in the sdk, ...

bwilkerson: > ... these are script references embedded in strings that are user facing ...

Oh, I missed that. Then yes, I agree, this should be happening on the server. I have to question why we're including a file reference in any of the messages, but if it's useful to the user then we should definitely translate it.

devoncarew: >  I have to question why we're including a file reference in any of the messages

Yeah, perhaps there's some other way to convey the info? Instead of `from dart:core/list.dart` perhaps, `from dart:core`?

bwilkerson: I need more context before I can form a concrete opinion / recommendation. What message / diagnostic are these file paths being included in?

test/utils_test.dart
johnpryan: I wasn't really either, I think this is used in the SDK.

bwilkerson: No, it isn't used in the sdk (there's no reference in `.dart_tool/package_config.json`). I'm not sure whether it's supported; it isn't configured to look like a dart-team supported package (https://pub.dev/packages/expected_output).

lib/src/utils.dart
johnpryan: As you point out, the full paths are being returned from the analyzer server. I created https://github.com/dart-lang/dart-services/pull/665 to address some of these issues. It now leaves `dart:core/foo.dart` alone. That way, we won't need to update anything if that's the format the analyzer chooses to use in the future.

lib/src/utils.dart
johnpryan: With https://github.com/dart-lang/dart-services/pull/665, `lib/core/foo.dart` will now be translated to `dart:core/foo.dart` and `lib/foo.dart` will be translated to `foo.dart`.

test/utils_test.unit
johnpryan: https://github.com/dart-lang/dart-services/pull/665 should be compatible with either format going forward.

PR #669, opened 2021-03-09
Copy Flutter analysis_options_user.yaml into flutter project dir.

gspencergoog:
# Description
Currently, the Flutter dartpad samples appear to be using the dart-services analysis_options.yaml file for their analysis options.  They should be using the analysis options in Flutter's `packages/flutter/analysis_options_user.yaml` file instead.  This PR adjusts grind.dart to do that.

# Related Issues

- https://github.com/flutter/flutter/issues/77425

/cc @domesticmouse

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -125,6 +125,14 @@ void buildProjectTemplates() async {
   final flutterProjectDir = await flutterProjectPath.create(recursive: true);
   joinFile(flutterProjectDir, ['pubspec.yaml'])
       .writeAsStringSync(createPubspec(includeFlutterWeb: true));
+  final optionsFile = await joinFile(Directory(flutterSdkPath),
+      ['packages', 'flutter', 'analysis_options_user.yaml']);
+  if (await optionsFile.exists()) {
+    optionsFile
+        .copy(path.join(flutterProjectDir.path, 'analysis_options.yaml'));
+  } else {
+    log("Flutter user analysis options file ${optionsFile.path} doesn't exist, using default.");
+  }
   await _runFlutterPubGet(flutterProjectDir);
 }
 
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -119,21 +119,21 @@ void buildProjectTemplates() async {
   joinFile(dartProjectDir, ['pubspec.yaml'])
       .writeAsStringSync(createPubspec(includeFlutterWeb: false));
   await _runDartPubGet(dartProjectDir);
+  joinFile(dartProjectDir, ['analysis_options.yaml'])
+      .writeAsStringSync(createDartAnalysisOptions());
 
   final flutterProjectPath =
       Directory(path.join(templatesPath.path, 'flutter_project'));
   final flutterProjectDir = await flutterProjectPath.create(recursive: true);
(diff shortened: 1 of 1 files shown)


tool/grind.dart
```diff
@@ -125,6 +125,14 @@ void buildProjectTemplates() async {
   final flutterProjectDir = await flutterProjectPath.create(recursive: true);
   joinFile(flutterProjectDir, ['pubspec.yaml'])
       .writeAsStringSync(createPubspec(includeFlutterWeb: true));
+  final optionsFile = await joinFile(Directory(flutterSdkPath),
```
devoncarew, line 128: I think here, we want to write a mostly empty analysis options file, probably w/ a comment, but that just references the flutter analysis options file (that will be resolvable because of the pubspec import below).

```yaml
include: package:flutter/analysis_options_user.yaml
```

devoncarew: And for the dart template project above, we'd just want an empty analysis options file so that the file for the dart-service project doesn't get picked up.

devoncarew (commented):
two drive by comments

tool/grind.dart
gspencergoog: Are there not some "basic" things we would want to include, even in the dart template project?  Like `unnecessary_new`, `unnecessary_const`, etc.?

devoncarew: It's more that they're tough to nail down. pedantic generally is a bit two proscriptive for casual dartpad samples for example.

Some small set seems reasonable, I'm just not sure where to draw the line for which to include.

The Dart team is working on developing a canonical recommended set of lints but that won't be ready for a little bit. A preview is here https://github.com/devoncarew/simple_lint_rules/blob/main/lib/analysis_options.yaml#L14, but I'm not sure we want to switch to that before all the rules are nailed down.

gspencergoog: Well, seeing as DartPad has been running with the rules from this repo until now, perhaps we should just use those as a starting point, and convert to the recommended set once it is available.

gspencergoog: Or we could just start with that list, and update it once it has been nailed down. That looks like a pretty decent starting point to me...

devoncarew: > Well, seeing as DartPad has been running with the rules from this repo until now, perhaps we should just use those as a starting point, and convert to the recommended set once it is available.

That seems reasonable to me (also happy w/ the potential canonical rule set; they're intended to be not surprising to users).

devoncarew (approved):
This looks good to me, but I'll ask somebody else from the dartpad team to take a quick look as well.

domesticmouse (approved):
LGTM. I'll roll it to production later today.

RedBrogdon:
A) Thanks for the quick review, @domesticmouse. 
B) Thanks for the PR, @gspencergoog. This is a great catch.
C) If I'm reading [this comment](https://github.com/dart-lang/dart-services/pull/669/files#diff-e8e8f48635252ccbc761873fcffe7b675f438e44a3533c65839a24a1aade7e9bR131) right, there's no functional change to the linters in place for our Dart or Flutter samples even after this is deployed, right? It looks like you're already thinking ahead to avoid breaking anything, but I wanted to check.

PR #676, opened 2021-03-18
Give the Dart Analysis Servers a config

domesticmouse:
Fix: https://github.com/dart-lang/dart-pad/issues/1784

### DIFF
--- a/benchmark/bench.dart
+++ b/benchmark/bench.dart
@@ -11,6 +11,7 @@ import 'package:dart_services/src/analysis_server.dart';
 import 'package:dart_services/src/bench.dart';
 import 'package:dart_services/src/common.dart';
 import 'package:dart_services/src/compiler.dart';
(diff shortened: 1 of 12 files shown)


srawlins (approved):
I'm not sure how the title matches the changes, but I like the changes!

domesticmouse:
@srawlins the core of this change is in `lib/src/analysis_server.dart`. Previously the Dart Analysis server was fed an empty directory as it's working directory, so it started enforcing null safety when we rolled over to dart 2.12. This change gives the Dart Analysis server a populated directory with a `pubspec.yaml` this is enough to stop the dart analysis server enforcing null safety rules.

Now if only I could figure out how to unbreak the Github Actions...

srawlins:
Ah I get it; makes sense. No ideas about the GitHub Actions :(

domesticmouse:
replaced by https://github.com/dart-lang/dart-services/pull/680

PR #683, opened 2021-04-01
Document protobuf installation a bit more

srawlins:


tool/grind.dart
```diff
@@ -312,10 +312,16 @@ void buildbot() => null;
 
 @Task('Generate Protobuf classes')
 void generateProtos() async {
-  await runWithLogging(
-    'protoc',
-    arguments: ['--dart_out=lib/src', 'protos/dart_services.proto'],
-  );
+  try {
```
domesticmouse, line 315: Can you push this `try`/`catch` into the `runWithLogging` function on https://github.com/srawlins/dart-services/blob/docs/tool/grind.dart#L342? 

We can add the informative error message as an optional parameter.

domesticmouse:
Also, you probably want to pull from head. You are trailing `master` by a commit, IIUC.

tool/grind.dart
srawlins: Done.

srawlins:
Rebased, thanks!

domesticmouse (approved):
Lgtm!

PR #689, opened 2021-04-06
Null safe and unsafe in a single codebase

domesticmouse:
Crunching both null safety and null unsafe into the same codebase to make the Cloud Run publishing cycle easier.

/cc @johnpryan this is the context for https://github.com/dart-lang/dart-pad/pull/1812 =)


lib/src/compiler.dart
```diff
@@ -150,12 +158,19 @@ class Compiler {
         '--modules=amd',
         if (usingFlutter) ...[
           '-s',
-          _flutterWebManager.summaryFilePath,
+          _flutterWebManager.summaryFilePath(_nullSafety),
           '-s',
-          '${Sdk.flutterBinPath}/cache/flutter_web_sdk/flutter_web_sdk/kernel/flutter_ddc_sdk.dill'
+          if (_nullSafety)
+            '${Sdk.flutterBinPath}/cache/flutter_web_sdk/flutter_web_sdk/kernel/flutter_ddc_sdk_sound.dill'
```
srawlins, line 164: Perhaps worth adding a field, `Sdk.flutterWebSdkKernelPath` or similar?

srawlins (approved):
Very clean. I only spot-reviewed `common.dart`. I like all of the grind changes.

lib/src/compiler.dart
domesticmouse: I'm curious if there will be usages apart from here for that method?

srawlins: Fair. The line is just so long 😛 .

domesticmouse: true dat =)

RedBrogdon (approved):
Is it weird that the most interesting part of this is the grinder file? 😄

lib/services_dev.dart
```diff
@@ -63,21 +62,19 @@ class EndpointsServer {
...
-  final String proxyTarget;
 
   Pipeline pipeline;
   Handler handler;
 
   CommonServerApi commonServerApi;
   FlutterWebManager flutterWebManager;
 
-  EndpointsServer._(this.port, this.proxyTarget) {
-    final commonServerImpl = (proxyTarget != null && proxyTarget.isNotEmpty)
```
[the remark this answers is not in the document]
domesticmouse: https://github.com/dart-lang/dart-services/pull/691 =)

lib/src/compiler.dart
domesticmouse: I'm not quite sure what a concurrent literal is, but how about a ternary operator to crunch out the common substring?

PR #692, opened 2021-04-08
Add a new `grind update-pub-dependencies` task to handle additional packages

srawlins:
The new task creates a scratch pub package using "any" version constraints on
Firebase packages. `pub get` then resolves the version constraints into exact
version numbers. We read the resolved version numbers and save them to
checked-in JSON files in the `tool/` directory.

The existing `buildProjectTemplates` and `buildStorageArtifacts` tasks then use
the date from those checked-in JSON files when writing pubspec.yaml files. This
keeps the versions known and checked into source control.

https://github.com/dart-lang/dart-pad/issues/1813


tool/grind.dart
```diff
@@ -455,3 +465,89 @@ linter:
...
+    },
+  );
+  joinFile(tempDir, ['pubspec.yaml']).writeAsStringSync(pubspec);
+  await _runFlutterPubGet(tempDir);
+  final pubspecLock =
+      loadYamlDocument(joinFile(tempDir, ['pubspec.lock']).readAsStringSync());
+  final pubSpecLockContents = pubspecLock.contents as YamlMap;
+  final packages = pubSpecLockContents['packages'] as YamlMap;
+  final packageVersions = <String, String>{};
+  final flutterPackages = [
```
parlough, line 509: Can this be `const`?

srawlins: Good catch. Done.

srawlins:
@domesticmouse some basic questions that I wasn't sure about:

* [ ] is `tool/*.json` a good place for these data files?
* [ ] is JSON a decent choice? I went through a few iterations:
  * Copy the pubspec.lock files directly - you have to open them up and parse them some time or another, so I figured might as well scrape out the version numbers early.
  * Write a valid Dart source file - This was looking nice as well, but it should be formatted to; in order to guarantee formatting, I'd have to bring in `dart_style`, which might bring a host of other packages, and really its just data. Should probably store it in a data format.
  * JSON is dead simple to encode and decode. YAML is harder, but does allow comments. I couldn't find an out-of-the-box JSON encoder which pretty-prints the JSON with newlines and indentations...
* [ ] Any issues with my Grinder tasks? Not sure how you like to organize those. I think it is correct that it depends on nothing? And nothing depends on it? It is a stand-alone task.
* [ ] OK to commit this to master without a flag? Nothing changes about what a user is allowed to import so should have no functional change.

domesticmouse (approved):
Lgtm!

domesticmouse:
> @domesticmouse some basic questions that I wasn't sure about:
> 
> Is `tool/*.json` a good place for these data files?

I'm fine with that. @devoncarew might have preferences tho?

> Is JSON a decent choice? I went through a few iterations:
>   
>   * Copy the pubspec.lock files directly - you have to open them up and parse them some time or another, so I figured might as well scrape out the version numbers early.
>   * Write a valid Dart source file - This was looking nice as well, but it should be formatted to; in order to guarantee formatting, I'd have to bring in `dart_style`, which might bring a host of other packages, and really its just data. Should probably store it in a data format.
>   * JSON is dead simple to encode and decode. YAML is harder, but does allow comments. I couldn't find an out-of-the-box JSON encoder which pretty-prints the JSON with newlines and indentations...

I'm happy with JSON. As for printing out JSON with newlines and indentations, I think [JsonEncoder.withIndent](https://api.dart.dev/stable/2.12.2/dart-convert/JsonEncoder/JsonEncoder.withIndent.html) is what you want.

> Any issues with my Grinder tasks? Not sure how you like to organize those. I think it is correct that it depends on nothing? And nothing depends on it? It is a stand-alone task.

I suspect it, at the very least, depends on the Flutter SDK being present.

> OK to commit this to master without a flag? Nothing changes about what a user is allowed to import so should have no functional change.

SGTM

srawlins:
> As for printing out JSON with newlines and indentations, I think JsonEncoder.withIndent is what you want.

Oh thank goodness. Done.

> I suspect it, at the very least, depends on the Flutter SDK being present.

Ah ha. Done.

tool/pub_dependencies_null-safe.json
```diff
@@ -0,0 +1,48 @@
+{
```
domesticmouse, line 1: I'm curious under what conditions the null safe and null unsafe dependencies may differ. They are currently identical AFAICT.

domesticmouse (approved):
Thanks for converting the dependency versions file to indented html. It's readable now =)

parlough (approved):

devoncarew:
Hey - apologies in advance - I  did not review this PR thoroughly. Two questions:

- when running pub get, do we run it in the context of the sdk referenced in https://github.com/dart-lang/dart-services/blob/master/flutter-sdk-version.yaml ? I'd want to make sure we resolve packages for the sdk version instead of something like the dart sdk which is driving grind / the services runtime
- is it necessary to check the package versions in? I suppose they'll be updated whenever we change flutter-sdk-version.yaml; do we want them to evolve faster? I guess, do we want to use the versions that solved when we last upgraded the base sdk, or, the latest published versions that solve against the sdk?

srawlins:
> when running pub get, do we run it in the context of the sdk referenced in https://github.com/dart-lang/dart-services/blob/master/flutter-sdk-version.yaml ?

Correct it is the Flutter SDK.

> is it necessary to check the package versions in?

I feel that it is important to fix package versions at a point in time that is controlled by the dart-services developer. This ensures that we spot-check version numbers, that version numbers are fixed ahead of deploy-time, and can be bumped separate from deploy-time.

One scenario I've tried to play out in my head is the case of a _bad_ release. Meaning a release that is either malicious or even just contains a bug, or a bug that specifically affects usage in DartPad. If we needed to rollback a dependency package, then source-controlling the version numbers gives a very straightforward way to do this: change the line in grind from `any` to a VersionConstraint which works around the malicious version, grind, deploy. Without the source control, you would still simply change the VersionConstraint from `any` to something more specific, and then deploy, but it is harder and takes longer to verify that the deployment has "pub gotten" good releases.

devoncarew:
> I feel that it is important to fix package versions at a point in time that is controlled by the dart-services developer. This ensures that we spot-check version numbers, that version numbers are fixed ahead of deploy-time, and can be bumped separate from deploy-time.

OK, that makes sense to me.

tool/pub_dependencies_null-safe.json
srawlins: Yeah I think they're identical. I'm also not clear about a good example of when they would be different. I think at some points during migration we would set the SDK constraints to something like `>=2.12.0-dev.0 <2.12.0`, but I don't think we're doing that any more.

The case is when a package somehow has an upper bound like that; a ceiling that excludes 2.12.0 or 2.13.0, but includes versions just prior. Kind of odd, but I think it's worth it to defensively store two lists. Probably a bad idea to run `pub get` with one pubspec with one SDK constraint, and apply the package resolution to a different pubspec.

PR #700, opened 2021-05-03
Add firebase packages to allowed import prefixes

srawlins:


srawlins:
@RedBrogdon @domesticmouse @johnpryan  this will effectively enable the feature, if dart-pad is deployed with the recent change there.

What should the general flow of testing / verification be? Wait to land this until dart-pad is deployed?

This change enables code which imports Firebase to be analyzed and compiled. The change landed in dart-pad allows code which imports Firebase to be executed with the right JS.

domesticmouse:
My feeling is this should land before the DartPad code because this is a
backwards compatible change.

My guess is landing DartPad first could confuse users who find the new UI
functionality which isn’t supported by the backend?

domesticmouse (approved):

RedBrogdon (approved):

PR #704, opened 2021-05-13
Add Pub package versions to VersionResponse

srawlins:
Deploy with https://github.com/dart-lang/dart-pad/pull/1887.

Simple change adds Pub package versions to VersionResponse, and fills in data with appropriate pubspec.lock info.


cloud_run.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM google/dart:2.12.4
+FROM google/dart:2.12.2
```
domesticmouse, line 1: We should think about removing the docker file update from the grind script.

lib/src/common_server_impl.dart
```diff
@@ -151,9 +152,11 @@ class CommonServerImpl {
               <String, String>{});
   }
 
-  Future<proto.VersionResponse> version(proto.VersionRequest _) {
+  Future<proto.VersionResponse> version(proto.VersionRequest request) {
```
domesticmouse, line 155: Is the `request` used?

domesticmouse (approved):

cloud_run.Dockerfile
domesticmouse: (revert this change)

lib/src/common_server_impl.dart
srawlins: Good catch, reverted.

cloud_run.Dockerfile
srawlins: done

PR #705, opened 2021-05-13
Migrate the example webapp to null safe

domesticmouse:


domesticmouse:
Looks like `example` has to be at the same null safety status as the main app. Yay.

PR #710, opened 2021-05-19
Allow pedantic to be imported

srawlins:


domesticmouse (approved):

srawlins:
I can wait until after I/O to land if its not important then.

PR #712, opened 2021-05-21
Use package lints

domesticmouse:
Hey Sam,

Is this a path we should go down?

brett

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -17,9 +17,12 @@ analyzer:
 
 linter:
   rules:
+    - always_declare_return_types
     - directives_ordering
     - prefer_final_fields
     - prefer_final_in_for_each
     - prefer_final_locals
     - prefer_relative_imports
+    - prefer_single_quotes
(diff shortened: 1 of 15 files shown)


srawlins:
I just made this change to two of my own repos 😁 .

When migrating from pedantic, there are 6 rules that are dropped, which we may want to manually list in analysis_options: https://github.com/dart-lang/lints/pull/41

The only interesting one is probably `always_declare_return_types`, maybe `prefer_single_quotes` and `unawaited_futures`.

domesticmouse:
> When migrating from pedantic, there are 6 rules that are dropped, which we may want to manually list in analysis_options: [dart-lang/lints#41](https://github.com/dart-lang/lints/pull/41)
> 
> The only interesting one is probably `always_declare_return_types`, maybe `prefer_single_quotes` and `unawaited_futures`.

I just added `always_declare_return_types`, `prefer_single_quotes` and `unawaited_futures` to `analysis_options.yaml`.

PR #706, opened 2021-05-13
Use pkg:lints and pkg:flutter_lints

domesticmouse:
fixes: https://github.com/dart-lang/dart-pad/issues/1881

PR #722, opened 2021-05-28
Use `pkg:lints` and `pkg:flutter_lints` re-build

domesticmouse:
Re-manufacturing #706, because reasons.

fixes: https://github.com/dart-lang/dart-pad/issues/1881


srawlins (approved):
Love it

lib/src/common.dart
```diff
@@ -189,11 +187,11 @@ class _MyHomePageState extends State<MyHomePage> {
         child: Column(
           mainAxisAlignment: MainAxisAlignment.center,
           children: <Widget>[
-            Text(
+            const Text(
               'You have pushed the button this many times:',
             ),
             Text(
-              '$_counter',
+              _counter.toString(),
```
mit-mit, line 194: I don't think the lints require this change. Isn't it better to keep this aligned with the `flutter create` template?

domesticmouse: Without that change the variable is marked as unused. I only made the changes required to get the tests to pass.

mit-mit: That sounds unexpected. Got a log or some repro steps for that issue?

mit-mit:
I landed my PR, so the diff here is now much smaller.

lib/src/common.dart
domesticmouse: Turns out I was holding it wrong, raw strings to the rescue.

mit-mit (approved):
Thanks Brett!

domesticmouse:
Blocked on https://github.com/flutter/codelabs/pull/134 and https://github.com/flutter/samples/pull/804

domesticmouse:
@RedBrogdon What else is required for this PR to land?

mit-mit:
@RedBrogdon @domesticmouse can we merge this now?

domesticmouse:
> @RedBrogdon @domesticmouse can we merge this now?

Hey @mit-mit, we are going to land this early next week. There will be some samples in various places that will have analysis warnings we this lands, and we want @RedBrogdon at the ready to push updates to the docs as required.

domesticmouse:
Landing this on master now. I'll switch over the servers tomorrow.

PR #724, opened 2021-06-02
Allow a few package imports in non-Flutter scripts

srawlins:
If a user script does not appear to be a Flutter script, our current pubspec templates do not allow _any_ package imports. This changes that to add a few specific packages.

This also tightens up the notions of allowed packages; previously we were relying on string implementations, which can lead to bugs (e.g. if a certain Sam forgets to allow flutter_test in writing `package:flutter/`).

Here's the list:

* characters
* collection
* js
* meta
* path
* pedantic
* vector_math

### DIFF
--- a/lib/src/project.dart
+++ b/lib/src/project.dart
@@ -29,15 +29,26 @@ String summaryFilePath(bool nullSafety) {
   );
 }
 
-const Set<String> _flutterImportPrefixes = {
-  'package:cloud_firestore/',
-  'package:firebase/',
-  'package:firebase_auth/',
-  'package:firebase_core/',
-  'package:flutter/',
-  'package:flutter_test/',
-  'package:pedantic/',
-  'dart:ui',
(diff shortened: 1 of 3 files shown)


srawlins:
cc @RedBrogdon @zoeyfan

domesticmouse:
@srawlins looks like you [broke the tests](https://github.com/dart-lang/dart-services/pull/724/checks#step:9:867) by not stopping imports =)

domesticmouse:
I'm unsure how, but it looks like this PR enables importing of `package:path`... which is not what I'd expect from this code. Any ideas on how it broke?

ps: you can replicate locally by running `grind test` =)

srawlins:
OK test is fixed; `package:path` is one of the allowed packages.

domesticmouse (approved):

srawlins:
This does not require any coordination with the front-end; deploying this will allow these packages to be directly imported.

PR #727, opened 2021-06-14
Bump caching of versioned artifacts to a week

domesticmouse:
Following the cache control configuration suggestion from:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#caching_static_assets

fixes: https://github.com/dart-lang/dart-pad/issues/1608


domesticmouse:
TBR: @devoncarew 

Landing this PR to include it in the artifacts publish for Flutter 2.2.2 / Dart 2.13.3

PR #729, opened 2021-06-17
Refactor stripFilePaths to better detect OK paths

srawlins:
Fixes https://github.com/dart-lang/dart-pad/issues/1920

* Remove dependency on expected_output; this package used mirrors, and our sole use was tests for stripFilePaths
* Rename stripFilePaths as it does not exclusively strip them
* Improve regex
* Improve performance;  static pattern, non-eager checks for different substrings


test/utils_test.dart
```diff
@@ -1,16 +1,61 @@
...
+  test('normalizeFilePaths keeps a "dart:core" path intact', () {
+    expectNormalizeFilePaths(
+      'dart:core/foo.dart',
+      'dart:core/foo.dart',
+    );
+  });
+
+  test('normalizeFilePaths keeps a web URL intact', () {
+    expectNormalizeFilePaths(
+      'See http://dart.dev/go/non-promo-property',
```
parlough, line 50: Great tests!

parlough (approved):
Thanks for working on this! Much better cleanup than the patch I initially worked on just for the URL case.

domesticmouse (approved):

PR #734, opened 2021-06-25
Migrate from dartis to resp_client

domesticmouse:


lib/src/server_cache.dart
```diff
@@ -99,30 +102,31 @@ class RedisCache implements ServerCache {
     if (_isShutdown) {
       return;
     }
-    log.info('$_logPrefix: reconnecting to $redisUriString...');
+    log.info('$_logPrefix: reconnecting to ${redisUri.toString()}...');
```
RedBrogdon, line 105: Is `toString` not called automatically by Dart's string interpolation?

RedBrogdon (approved):
LGTM, assuming you've given the `resp_client` itself a good looking over.

What's the best way to test something like this as it's deployed, since if there's a weird bug it could potentially pollute our prod Memorystore?

lib/src/server_cache.dart
domesticmouse: Heh, forrest for the trees. Fixed =)

domesticmouse:
> LGTM, assuming you've given the `resp_client` itself a good looking over.

Yup, I've read through the `resp_client` codebase, it is a lot easier to understand than the type dispatch magic at the core of `dartis`.

> What's the best way to test something like this as it's deployed, since if there's a weird bug it could potentially pollute our prod Memorystore?

We could add something to the generated keys such that the `dartis` cached documents and the `resp_client` cached documents are in disjoint keyspaces.

PR #735, opened 2021-06-27
Null safety migration

domesticmouse:
fixes: https://github.com/dart-lang/dart-pad/issues/1796

### DIFF
--- a/benchmark/bench.dart
+++ b/benchmark/bench.dart
@@ -53,7 +53,7 @@ void main(List<String> args) async {
 
 class AnalyzerBenchmark extends Benchmark {
   final String source;
-  AnalysisServerWrapper analysisServer;
+  late AnalysisServerWrapper analysisServer;
 
   AnalyzerBenchmark(
     String name,
@@ -63,7 +63,7 @@ class AnalyzerBenchmark extends Benchmark {
   }
 
   @override
-  Future<AnalysisServer> init() => analysisServer.init();
+  Future<AnalysisServer?> init() => analysisServer.init();
 
   @override
   Future<proto.AnalysisResults> perform() => analysisServer.analyze(source);
@@ -111,7 +111,7 @@ class AnalysisServerBenchmark extends Benchmark {
         super('completion.$name');
 
   @override
-  Future<AnalysisServer> init() => analysisServer.init();
+  Future<AnalysisServer?> init() => analysisServer.init();
 
   @override
   Future<proto.CompleteResponse> perform() =>
--- a/benchmark/bench.dart
+++ b/benchmark/bench.dart
@@ -53,14 +53,13 @@ void main(List<String> args) async {
 
 class AnalyzerBenchmark extends Benchmark {
   final String source;
-  late AnalysisServerWrapper analysisServer;
+  final AnalysisServerWrapper analysisServer;
 
   AnalyzerBenchmark(
     String name,
     this.source,
-  ) : super('analyzer.$name') {
-    analysisServer = DartAnalysisServerWrapper(nullSafe);
-  }
+  )   : analysisServer = DartAnalysisServerWrapper(nullSafe),
+        super('analyzer.$name');
 
   @override
   Future<AnalysisServer?> init() => analysisServer.init();
--- a/example/web/index.dart
+++ b/example/web/index.dart
@@ -13,7 +13,7 @@ import 'services_utils.dart' as utils;
 
 const versions = ['v2'];
 
-BrowserClient _client;
+BrowserClient? _client;
 
 void main() {
   setupAnalyze();
@@ -26,9 +26,9 @@ void main() {
   setupVersion();
 }
 
-Future<Response> post(String url, {String body}) async {
+Future<Response> post(String url, {String? body}) async {
   _client ??= utils.SanitizingBrowserClient();
-  return _client.post(
+  return _client!.post(
     Uri.parse(url),
     body: body,
     encoding: utf8,
@@ -38,37 +38,37 @@ Future<Response> post(String url, {String body}) async {
 
 Future<Response> get(String url) async {
   _client ??= utils.SanitizingBrowserClient();
-  return _client.get(
+  return _client!.get(
     Uri.parse(url),
     headers: {'content-type': 'application/json'},
   );
 }
 
 void setupAnalyze() {
-  versions.forEach((version) {
+  for (final version in versions) {
     final editor =
-        createEditor(querySelector('#analyzeSection-$version .editor'));
+        createEditor(querySelector('#analyzeSection-$version .editor')!);
     final output = querySelector('#analyzeSection-$version .output');
     final button =
         querySelector('#analyzeSection-$version button') as ButtonElement;
     button.onClick.listen((e) {
-      final source = {'source': editor.getDoc().getValue()};
+      final source = {'source': editor.getDoc()!.getValue()};
       final sw = Stopwatch()..start();
       post(
         '$_uriBase/dartservices/$version/analyze',
         body: json.encode(source),
       ).then(
-          (response) => output.text = '${_formatTiming(sw)}${response.body}');
+          (response) => output!.text = '${_formatTiming(sw)}${response.body}');
     });
-  });
+  }
 }
 
 void setupAssists() {
-  versions.forEach((version) {
+  for (final version in versions) {
     final editor =
-        createEditor(querySelector('#assistsSection-$version .editor'));
+        createEditor(querySelector('#assistsSection-$version .editor')!);
     final output = querySelector('#assistsSection-$version .output');
-    final offsetElement = querySelector('#assistsSection-$version .offset');
+    final offsetElement = querySelector('#assistsSection-$version .offset')!;
     final button =
         querySelector('#assistsSection-$version button') as ButtonElement;
     button.onClick.listen((e) {
@@ -78,61 +78,61 @@ void setupAssists() {
         '$_uriBase/dartservices/$version/assists',
         body: json.encode(source),
       ).then(
-          (response) => output.text = '${_formatTiming(sw)}${response.body}');
+          (response) => output!.text = '${_formatTiming(sw)}${response.body}');
     });
     offsetElement.text = 'offset ${_getOffset(editor)}';
     editor.onCursorActivity.listen((_) {
       offsetElement.text = 'offset ${_getOffset(editor)}';
     });
-  });
+  }
 }
 
 void setupCompile() {
-  versions.forEach((version) {
+  for (final version in versions) {
     final editor =
-        createEditor(querySelector('#compileSection-$version .editor'));
+        createEditor(querySelector('#compileSection-$version .editor')!);
     final output = querySelector('#compileSection-$version .output');
     final button =
         querySelector('#compileSection-$version button') as ButtonElement;
     button.onClick.listen((e) {
-      final source = editor.getDoc().getValue();
+      final source = editor.getDoc()!.getValue();
       final compile = {'source': source};
       final sw = Stopwatch()..start();
       post(
         '$_uriBase/dartservices/$version/compile',
         body: json.encode(compile),
       ).then(
-          (response) => output.text = '${_formatTiming(sw)}${response.body}');
+          (response) => output!.text = '${_formatTiming(sw)}${response.body}');
     });
-  });
+  }
 }
 
 void setupCompileDDC() {
-  versions.forEach((version) {
+  for (final version in versions) {
     final editor =
-        createEditor(querySelector('#compileDDCSection-$version .editor'));
+        createEditor(querySelector('#compileDDCSection-$version .editor')!);
     final output = querySelector('#compileDDCSection-$version .output');
     final button =
         querySelector('#compileDDCSection-$version button') as ButtonElement;
     button.onClick.listen((e) {
-      final source = editor.getDoc().getValue();
+      final source = editor.getDoc()!.getValue();
       final compile = {'source': source};
       final sw = Stopwatch()..start();
       post(
         '$_uriBase/dartservices/$version/compileDDC',
         body: json.encode(compile),
       ).then(
-          (response) => output.text = '${_formatTiming(sw)}${response.body}');
+          (response) => output!.text = '${_formatTiming(sw)}${response.body}');
     });
-  });
+  }
 }
 
 void setupComplete() {
-  versions.forEach((version) {
+  for (final version in versions) {
     final editor =
-        createEditor(querySelector('#completeSection-$version .editor'));
+        createEditor(querySelector('#completeSection-$version .editor')!);
     final output = querySelector('#completeSection-$version .output');
-    final offsetElement = querySelector('#completeSection-$version .offset');
+    final offsetElement = querySelector('#completeSection-$version .offset')!;
     final button =
         querySelector('#completeSection-$version button') as ButtonElement;
     button.onClick.listen((e) {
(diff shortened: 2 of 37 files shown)


domesticmouse:
This is a big enough change that I feel it deserves multiple pairs of eyes

lib/src/server_cache.dart
```diff
@@ -108,7 +108,7 @@ class RedisCache implements ServerCache {
       // 1 <= (randomSource.nextDouble() + 1) < 2
       nextRetryMs = (retryTimeoutMs * (randomSource.nextDouble() + 1)).toInt();
     }
-    connectSocket(redisUri.host, port: redisUri.hasPort ? redisUri.port : null)
+    connectSocket(redisUri.host, port: redisUri.hasPort ? redisUri.port : 6379)
```
RedBrogdon, line 111: Is this the default port used by Cloud Memorystore?

RedBrogdon (approved):
This LGTM, but I think you're right to want a couple different reviewers.

lib/services_cloud_run.dart
```diff
@@ -79,22 +79,22 @@ class EndpointsServer {
...
+  late CommonServerImpl _commonServerImpl;
 
   EndpointsServer._(this.port, this.redisServerUri, bool nullSafety) {
     _commonServerImpl = CommonServerImpl(
       _ServerContainer(),
       redisServerUri == null
           ? InMemoryCache()
           : RedisCache(
-              redisServerUri,
+              redisServerUri!,
```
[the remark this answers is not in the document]
parlough: Since the constructor uses the constructor field initializer for this, `redisServerUri` is technically a property(`this.redisServerUri`) and [cannot be promoted](https://dart.dev/tools/non-promotion-reasons).

lib/src/analysis_server.dart
```diff
@@ -234,31 +234,45 @@ abstract class AnalysisServerWrapper {
       await _loadSources(<String, String>{mainPath: source});
       await analysisCompleter.future;
 
-      final result = await analysisServer.analysis.getHover(mainPath, offset);
+      final result = await analysisServer!.analysis.getHover(mainPath, offset);
       await _unloadSources();
 
       if (result.hovers.isEmpty) {
-        return null;
+        return <String, String>{};
```
parlough, line 241: Is this map added to later on? Perhaps this could be made const?

```suggestion
        return const {};
```

lib/src/analysis_server.dart
```diff
@@ -234,31 +234,45 @@ abstract class AnalysisServerWrapper {
...
+        m['deprecated'] = '${info.isDeprecated}';
+      }
 
-      for (final key in m.keys.toList()) {
-        if (m[key] == null) m.remove(key);
+      if (info.staticType != null) {
+        m['staticType'] = info.staticType!;
+      }
+      if (info.propagatedType != null) {
+        m['propagatedType'] = info.propagatedType!;
```
parlough, line 275: Perhaps this would be a good candidate for a collection if?

```dart
return {
    if (info.elementDescription != null)
      'description': info.elementDescription!,
    if (info.elementKind != null) 'kind': info.elementKind!,
    if (info.dartdoc != null) 'dartdoc': info.dartdoc!,
    ...
  };
```

lib/src/analysis_server.dart
```diff
@@ -286,7 +300,7 @@ abstract class AnalysisServerWrapper {
           ..line = error.location.startLine
           ..message = utils.normalizeFilePaths(error.message)
           ..sourceName = path.basename(error.location.file)
-          ..hasFixes = error.hasFix
+          ..hasFixes = error.hasFix!
```
parlough, line 303: If this property is nullable, are there any cases where this may be null that may occur in this case?

Perhaps a default value of `false` wouldn't hurt?

```suggestion
          ..hasFixes = error.hasFix ?? false
```

lib/src/analysis_server.dart
parlough, line 241: Will this returned map ever be modified? If not, could likely make this `const`:

parlough, line 241: Will this returned map ever be modified? If not, could likely make this `const`:

parlough, line 241: Will this map ever be modified? If not, could perhaps make this `const`:

```suggestion
        return const {};
```

lib/src/project.dart
```diff
@@ -90,11 +90,11 @@ List<ImportDirective> getUnsupportedImports(List<ImportDirective> imports) {
   return imports.where((import) {
     final uriString = import.uri.stringValue;
     // All non-VM 'dart:' imports are ok.
-    if (uriString.startsWith('dart:')) {
+    if (uriString!.startsWith('dart:')) {
       return !_allowedDartImports.contains(uriString);
     }
 
-    final uri = Uri.tryParse(import.uri.stringValue);
+    final uri = Uri.tryParse(import.uri.stringValue!);
```
parlough, line 97: Since you pull out `import.uri.stringValue` into its own local variable, you could add a null check to promote it and reuse the local variable instead of the field while avoiding the null assertions:

```dart
final uriString = import.uri.stringValue;
if (uriString == null) {
  return false;
}
// All non-VM 'dart:' imports are ok.
if (uriString.startsWith('dart:')) {
  return !_allowedDartImports.contains(uriString);
}

final uri = Uri.tryParse(uriString);
if (uri == null) return false;
```

lib/src/pub.dart
```diff
@@ -29,11 +28,11 @@ const _flutterPackages = [
 
 /// Each of these is expensive to calculate; they require reading from disk.
 /// None of them changes during execution.
-/*late*/ Map<String, String> _nullSafePackageVersions;
-/*late*/ Map<String, String> _preNullSafePackageVersions;
+Map<String, String>? _nullSafePackageVersions;
+Map<String, String>? _preNullSafePackageVersions;
```
parlough, line 32: Could you instead make this lazily initialized and then simplify the `getPackageVersions` method?

```suggestion
late final Map<String, String> _nullSafePackageVersions = packageVersionsFromPubspecLock(project.flutterTemplateProject(true));
late final Map<String, String> _preNullSafePackageVersions = packageVersionsFromPubspecLock(project.flutterTemplateProject(false));
```

Then turn the `getPackageVersions` method into the following:

```dart
/// Returns a mapping of Pub package name to package version.
Map<String, String> getPackageVersions({required bool nullSafe}) =>
    nullSafe ? _nullSafePackageVersions : _preNullSafePackageVersions;
```

parlough (commented):
This is nice to see. A few potential suggestions:

lib/src/analysis_server.dart
```diff
@@ -54,18 +54,18 @@ abstract class AnalysisServerWrapper {
   final String sdkPath;
   final TaskScheduler serverScheduler = TaskScheduler();
 
-  Future<AnalysisServer> _init;
+  Future<AnalysisServer?>? _init;
 
   /// Instance to handle communication with the server.
-  AnalysisServer analysisServer;
+  AnalysisServer? analysisServer;
```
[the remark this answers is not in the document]
domesticmouse: We are using null as a signal that the analysis server isn't running. But the health check that used this is no longer called. So I'm going with just throwing an error, which I believe will take the server down. Which is appropriate for Cloud Run, it'll get replaced by another instance by the Cloud Run infra.

lib/services_cloud_run.dart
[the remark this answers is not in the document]
domesticmouse: Converted `redisServerUri` to a parameter so it can be promoted

lib/src/server_cache.dart
domesticmouse: For some reason `Uri` is defined to return 0 for an unspecified port in a uri when the scheme isn't http or https. It should really be optional and return null in that case. Also `connectSocket` is defined as having an optional non nullable argument port. Which I don't want set to 0. We actually set the port in the `cloud_run.Dockerfile` redis url argument, so this is me just trying to idiot proof an unused code path.

And yes, redis' default connection port is 6379.

lib/src/analysis_server.dart
domesticmouse: Good call, done

lib/src/project.dart
domesticmouse: Good call, again =)

lib/src/pub.dart
domesticmouse: I've gone with your defaulting to null safety approach. Much betterer =)

domesticmouse:
> This LGTM, but I think you're right to want a couple different reviewers.

I'm holding off landing this until @srawlins has had a chance to review. I don't believe there is any rush on getting this change into production.

benchmark/bench.dart
```diff
@@ -53,7 +53,7 @@ void main(List<String> args) async {
 
 class AnalyzerBenchmark extends Benchmark {
   final String source;
-  AnalysisServerWrapper analysisServer;
+  late AnalysisServerWrapper analysisServer;
```
srawlins, line 56: It looks like this can be final and non-late? If it was assigned on line ~60 in the initializers? A TODO would be fine if the answer is "maybe" 😛

example/web/index.dart
```diff
@@ -24,9 +24,9 @@ void main() {
   setupVersion();
 }
 
-Future<Response> post(String url, {String body}) async {
+Future<Response> post(String url, {String? body}) async {
   _client ??= utils.SanitizingBrowserClient();
```
srawlins, line 28: I think what we want here is to change `_client` to be a late initialized field:

```dart
late BrowserClient _client = utils.SanitizingBrowserClient();
```

Making it late means the right side of `=` will not be executed until the first time it is referenced.

example/web/index.dart
```diff
@@ -76,47 +76,47 @@ void setupAssists() {
...
   button.onClick.listen((e) {
-    final source = editor.getDoc().getValue();
+    final source = editor.getDoc()!.getValue();
     final compile = {'source': source};
     final sw = Stopwatch()..start();
     post(
       '$_uriBase/dartservices/v2/compile',
       body: json.encode(compile),
-    ).then((response) => output.text = '${_formatTiming(sw)}${response.body}');
+    ).then((response) => output!.text = '${_formatTiming(sw)}${response.body}');
```
srawlins, line 89: just a style note: you might move this `!` to the end of line 80, so that `output` is definitely an element, and we fail faster.

example/web/index.dart
```diff
@@ -144,17 +144,17 @@ void setupDocument() {
 }
 
 void setupFixes() {
-  final editor = createEditor(querySelector('#fixesSection-v2 .editor'));
+  final editor = createEditor(querySelector('#fixesSection-v2 .editor')!);
   final output = querySelector('#fixesSection-v2 .output');
```
srawlins, line 148: Same here. Putting a `!` here means fewer `!` below.

lib/services_cloud_run.dart
```diff
@@ -79,16 +79,15 @@ class EndpointsServer {
   }
 
   final int port;
-  HttpServer server;
-  String redisServerUri;
+  late HttpServer server;
 
-  Pipeline pipeline;
-  Handler handler;
+  late Pipeline pipeline;
```
srawlins, line 84: I think these two can now be `late final`? I could be wrong.

lib/services_cloud_run.dart
```diff
@@ -79,16 +79,15 @@ class EndpointsServer {
...
+  late Handler handler;
 
-  CommonServerApi commonServerApi;
-  CommonServerImpl _commonServerImpl;
+  late CommonServerApi commonServerApi;
```
srawlins, line 87: I think this can also be `late final`.

lib/services_cloud_run.dart
```diff
@@ -79,16 +79,15 @@ class EndpointsServer {
   }
 
   final int port;
-  HttpServer server;
-  String redisServerUri;
```
srawlins, line 83: +1; nice removal

lib/services_dev.dart
```diff
@@ -60,12 +60,12 @@ class EndpointsServer {
   }
 
   final int port;
-  HttpServer server;
+  late HttpServer server;
```
srawlins, line 63: I can't find any usage of this. I might be crazy.

lib/services_dev.dart
```diff
@@ -60,12 +60,12 @@ class EndpointsServer {
...
 
-  Pipeline pipeline;
-  Handler handler;
+  late Pipeline pipeline;
```
srawlins, line 65: I think each of these three can be `late final`.

lib/src/analysis_server.dart
```diff
@@ -238,30 +238,25 @@ abstract class AnalysisServerWrapper {
...
-
-      m['staticType'] = info.staticType;
-      m['propagatedType'] = info.propagatedType;
-
-      for (final key in m.keys.toList()) {
-        if (m[key] == null) m.remove(key);
-      }
-
-      return m;
+      return {
```
srawlins, line 246: ❤️

lib/src/bench.dart
```diff
@@ -84,7 +84,7 @@ class BenchmarkHarness {
 
   Future<BenchMarkResult> _time(
       Benchmark benchmark, int minIterations, int minMillis,
-      [int maxMillis]) {
+      [int? maxMillis]) {
```
srawlins, line 87: Given just the logic below at line 97, could we make this non-nullable, defaulting to -1? I think that matches the current logic...

lib/src/sdk.dart
```diff
@@ -6,12 +6,11 @@ import 'dart:async';
 import 'dart:convert' show utf8;
 import 'dart:io';
 
-import 'package:meta/meta.dart';
 import 'package:path/path.dart' as path;
 import 'package:yaml/yaml.dart';
 
 class Sdk {
-  static Sdk _instance;
+  static Sdk? _instance;
```
srawlins, line 13: I think `static late final` (or however that is legal) would be best, and then scrap the factory constructor,  but that sounds like a big change so a TODO would be great.

srawlins (approved):
I only spot-checked the tests, but this looks great!

benchmark/bench.dart
domesticmouse: done

example/web/index.dart
domesticmouse: done

example/web/index.dart
domesticmouse: done

example/web/index.dart
domesticmouse: done

lib/services_cloud_run.dart
domesticmouse: done

lib/services_cloud_run.dart
domesticmouse: and done

lib/services_cloud_run.dart
domesticmouse: =)

lib/services_dev.dart
domesticmouse: Not crazy, you are just seeing the remains of messy evolution. Tidied up.

lib/services_dev.dart
domesticmouse: done

lib/src/analysis_server.dart
domesticmouse: That's @parlough's fine work right there =)

lib/src/bench.dart
domesticmouse: My one concern here is that IIUC all the benchmark code is currently not utilised, and not tested. I've left it alone in case someone wants to revive this functionality, but I'm increasingly leaning towards deleting it outright. Unused code is tech debt, and all that.

Also I'm unsure if `-1` is actually the correct default. My reading is that it should be either max int or some reasonable like `10 * 60 * 1000`.

lib/src/sdk.dart
domesticmouse: I've turned `_instance` into a `static late final` but I've put a TODO on the factory constructor.

lib/src/bench.dart
srawlins: Haha glad I asked. No worries here then. Not in the hot path of actual DartPad.

PR #737, opened 2021-06-27
Enable asserts

domesticmouse:
Hey @devoncarew,

Is this all that is required to enable asserts? Or am I holding it wrong?

brett

fixes: https://github.com/dart-lang/dart-pad/issues/310


devoncarew:
Hmm, good question; this is for dart2js and DDC? cc @sigmundch

domesticmouse:
Yup, I've added it on both the `compile` and `compileDDC` code paths.

lib/src/compiler.dart
```diff
@@ -70,6 +70,7 @@ class Compiler {
           '--sound-null-safety',
           '--enable-experiment=non-nullable',
```
sigmundch, line 71: unrelated: note that this flag is no longer needed (here and below in the call to `dartdevc`) now that null safety is enabled by default.

sigmundch (commented):
Yes, I believe that's all there is to it. I added a comment on the bug about whether we want to enable it unconditionally though.

devoncarew:
Hmm, from @sigmundch's comments here: https://github.com/dart-lang/dart-pad/issues/310#issuecomment-869828215 - I am wondering whether we actually do want to enable asserts. This will definitely have performance implications for flutter snippets - Flutter makes heavy use of asserts (only enabled in their debug mode for flutter mobile) in order to validate framework assumptions.

domesticmouse:
/cc @RedBrogdon for his thoughts.

Making this a choice visible in the front end ups the complexity of the solution significantly.

RedBrogdon:
Let's cover it in the next DartPad weekly. Feel free to add it to the agenda.

domesticmouse:
> Feel free to add it to the agenda.

Added

lib/src/compiler.dart
domesticmouse: Fix for this is in https://github.com/dart-lang/dart-services/pull/738

domesticmouse:
Marking as draft until we decide whether we want to land this as-is, or if it requires further work in the front end first.

domesticmouse:
@RedBrogdon & @srawlins I'm guessing I should close out this PR

RedBrogdon:
@domesticmouse and @johnpryan, is there any coordination needed between you two for this to go live? In other words, wold it be worth having some time to double-check that the front-end can properly catch and surface failed asserts before turning them on in the back end?

johnpryan:
Last week we looked into this, and with DDC compiled code, asserts are enabled by default, so `--enable-asserts` is a no-op but doesn't hurt. Dart2js can be slower with asserts enabled, but I still think we should turn them on for DartPad, since since I've been working with this branch locally. We can always turn them off again if we see performance issues.

johnpryan:
@RedBrogdon to answer your question, no coordination is needed for this to go live.

RedBrogdon (approved):

domesticmouse:
QQ: do we want to land this PR? I'm confused as to where we are at with this change =)

johnpryan:
I would like to go forward with this change, I think it will help display asserts in dart2js snippets

domesticmouse:
I'll land after the firebase event.

/cc @srawlins

PR #740, opened 2021-06-30
Refactor AnalysisServerWrapper with simpler init

srawlins:
Simplify AnalysisServerWrapper a bit:

* `_init` is unused except for "has it been assigned yet" so changed to `bool`.
* `init`'s return value is never used, so instead make it return void.

### DIFF
--- a/benchmark/bench.dart
+++ b/benchmark/bench.dart
@@ -53,17 +53,16 @@ void main(List<String> args) async {
 
 class AnalyzerBenchmark extends Benchmark {
   final String source;
-  AnalysisServerWrapper analysisServer;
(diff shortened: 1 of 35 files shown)


domesticmouse (approved):

domesticmouse:
Looks like this code needs to be fixed: https://github.com/dart-lang/dart-services/runs/2949223890#step:7:8

srawlins:
Thanks!

domesticmouse:
Any reason not to land this PR?

srawlins:
Sorry, was on vacation; thanks!

PR #749, opened 2021-09-16
Add PackageInfo proto with more info

srawlins:
Mirrored here: https://github.com/dart-lang/dart-pad/pull/2009

### DIFF
--- a/lib/src/common_server_impl.dart
+++ b/lib/src/common_server_impl.dart
@@ -8,7 +8,6 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:crypto/crypto.dart';
-import 'package:dart_services/src/project.dart';
 import 'package:logging/logging.dart';
 import 'package:pedantic/pedantic.dart';
 
@@ -16,6 +15,7 @@ import '../version.dart';
 import 'analysis_servers.dart';
 import 'common.dart';
 import 'compiler.dart';
+import 'project.dart';
 import 'protos/dart_services.pb.dart' as proto;
 import 'pub.dart';
 import 'sdk.dart';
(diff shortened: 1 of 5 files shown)


lib/src/common_server_impl.dart
```diff
@@ -153,6 +154,13 @@ class CommonServerImpl {
   Future<proto.VersionResponse> version(proto.VersionRequest _) {
     final sdk = Sdk.create();
     final packageVersions = getPackageVersions(nullSafe: _nullSafety);
+    final packageInfos = [
+      for (var packageName in packageVersions.keys)
+        proto.PackageInfo()
+          ..name = packageName
+          ..version = packageVersions[packageName]!
+          ..supported = isSupportedPackage(packageName),
```
parlough, line 162: A nit, but could you use `.entries` here to avoid the `!`?

```suggestion
      for (var package in packageVersions.entries)
        proto.PackageInfo()
          ..name = package.key
          ..version = package.value
          ..supported = isSupportedPackage(package.key),
```

srawlins: I hate losing the semantic meanings when using `entries`. :( I think it really is less readable coming back as a reader. "What is the 'key' of a package?" "What is the 'value' of a package?"

parlough: That's fair and the `!` isn't an issue here. Thanks for the clarification of your thoughts :)

parlough: I tried to resolve the original message, but deleted it...

parlough (approved):
This looks good to me, outside of the `import 'package:dart_services/src/project.dart';` causing analysis to fail.

PR #753, opened 2021-09-25
Use the dart docker image

domesticmouse:

PR #754, opened 2021-09-25
Revert "Use the dart docker image"

domesticmouse:
Reverts dart-lang/dart-services#753


domesticmouse:
TBR: @RedBrogdon 

As expected, moving from `google/dart` to `dart` fails on GCP Cloud Build. Raised https://github.com/dart-lang/dart-pad/issues/2015 to track this required migration effort.

PR #755, opened 2021-09-27
Support Flutter Plugins via a new project template

srawlins:
Fixes https://github.com/dart-lang/dart-pad/issues/2010

This PR grew and grew, but it takes quite a few adjustments to support Flutter plugins like `url_launcher`:

Flutter Web "registers" plugins during `flutter packages get` by writing a `lib/generated_plugin_registrant.dart` file, if your project has a `web/index.html` file. This is what wires up functions like `url_launcher`'s `canLaunch`. So I updated `tool/grind.dart` to use `flutter packages get` and write a `web/index.html`.

However, this now also registers some Firebase plugins in ways that weren't happening before, which now _expect_ a Firebase app on launch. The registration _requires_ a Firebase app. So if you are just writing a Flutter Web app, the JavaScript will crash if you do not have a Firebase app.

So I created a third project template directory: `firebase_project`, which is the same as the previous `flutter_project`. And now the `flutter_project` template is generated without Firebase packages in pubspec.yaml, so they are also not registered in `lib/generated_plugin_registrant.dart`.

In order to simplify all of this, I refactored `project.dart` to have a class, ProjectTemplates, which replaces most of the top-level functions that previously comprised that library.


domesticmouse (approved):

domesticmouse:
This PR is triggering failures. I'm unsure what is going wrong here, but I'd suggest we need green CI before landing this PR.

srawlins:
@domesticmouse any idea what's going on with the stable and beta tests w.r.t. flutter_lints? It passes for me locally; seems like a pretty odd state to be in, not reporting that lint...

domesticmouse:
I've rerun the CI and I get different passes and fails each time. Something is flaky.

domesticmouse:
Oh look, all the tests passed. After repeated prodding. Weird.

srawlins:
Accidental roll of the dice where all three pass? 😟

domesticmouse:
> Accidental roll of the dice where all three pass? 😟

Yeah, probably. Le sigh

domesticmouse:
Feel free to land when evs

PR #762, opened 2021-10-04
Add hooks_flutter and hooks_riverpod

srawlins:
Also restructure the multiple lists of supported packages into a single source of truth in `lib/src/project.dart`.


lib/src/project.dart
```diff
@@ -83,7 +92,6 @@ const Set<String> supportedNonFlutterPackages = {
   'http',
   'intl',
   'js',
-  'lints',
```
domesticmouse, line 86: I'm curious why this is being removed? I think this is required for the lints to reported for pure dart dartpads.

domesticmouse (approved):

lib/src/project.dart
srawlins: I was curious why this was here, since this is for import statements in the user code. `lints` and `flutter_lints` are still in the pubspec file, in `tool/grind.dart` lines 521 and 522.

domesticmouse: Hmm, yeah this shouldn't be needed. But the tests are failing. ugh

domesticmouse:
Updated to head, which is re-running CI as a by-product.

srawlins:
Triple red again? Maybe there is something wrong with this PR. #764 may reveal more.

PR #765, opened 2021-10-07
Bump analyzer dependency to 2.5.0

srawlins:
I'm not seeing flakes (as much?) locally, and my pubspec.lock is using analyzer 2.5.0, so I figure maybe we should bump it; perhaps past a bug.


domesticmouse (approved):

domesticmouse:
I'd be happy to pin all the `pubspec.yaml` min values at same as `pubspec.lock`.

PR #766, opened 2021-10-09
Do not register angry Firebase packages

srawlins:
I've identified some angry Firebase packages https://github.com/dart-lang/dart-pad/issues/2033, which require a Firebase app to be configured in JS, before Dart is executed, or they throw.

The workaround (very hacky) is to run `flutter packages get` with the safe Firebase packages. This ensures that plugins-which-must-be-registered, like url_launcher, are registered, and the angry Firebase ones are not. A second step is added, where we run `dart pub get` with the total set of Firebase packages, so that they are available (but not registered in Flutter's generated registrant file).

Fixes #2033


srawlins:
Failures are same old flakes.

domesticmouse (approved):

PR #768, opened 2021-10-11
Flutter 2.6.0-5.2.pre

domesticmouse:
Hey @srawlins, it looks like master is broken on Flutter 2.6 beta. Any ideas on what we need to do to fix this greatly appreciated.


domesticmouse:
Ok that's weird. It fails locally on macOS but succeeds in CI?!

PR #771, opened 2021-10-14
Change how compilers are invoked to prepare for upcoming deprecation.

sigmundch:
Change how compilers are invoked to prepare for upcoming deprecation of the dart2js and dartdevc scripts.

These scripts will be deprecated (possibly in 2.15 or later) and then removed in a future version of the Dart SDK (at least post 2.16). The `dartdevc` script will be removed, since it was never meant for end-users to invoke it, however the tool will continue to be available directly from the snapshot for dartpad and flutter tools. The `dart2js` script will be replaced by the "dart compile js" command (which is already available).

### DIFF
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -24,23 +24,20 @@ Logger _logger = Logger('compiler');
 class Compiler {
   final Sdk _sdk;
   final String _dartPath;
-  final String _dartdevcPath;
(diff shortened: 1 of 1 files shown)


domesticmouse (approved):
Lgtm once CI passes

domesticmouse:
Thanks @sigmundch, I'll land this shortly.

PR #786, opened 2021-11-04
Use a local pub cache

srawlins:
This puts all dependencies in a fix location in the local directory, and is necessary if we want to patch a dependency. We want to patch some firebase packages so that they work without JS Firebase config.

Work towards: https://github.com/dart-lang/dart-pad/issues/2068

The PUB_CACHE variable only needs to be used for `pub get` and `flutter packages get`, because after those run, the `.dart_tool/package_config.json` file has all absolute paths set, and no further tasks need to know about the custom PUB_CACHE.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -12,7 +12,6 @@ import 'package:dart_services/src/project.dart';
 import 'package:dart_services/src/pub.dart';
 import 'package:dart_services/src/sdk.dart';
 import 'package:grinder/grinder.dart';
-import 'package:grinder/src/run_utils.dart' show mergeWorkingDirectory;
 import 'package:http/http.dart' as http;
 import 'package:path/path.dart' as path;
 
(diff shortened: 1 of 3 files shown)


tool/grind.dart
```diff
@@ -280,26 +278,31 @@ linter:
...
     flutterToolPath,
     arguments: ['packages', 'get'],
     workingDirectory: dir.path,
+    environment: {'PUB_CACHE': _pubCachePath},
   );
 }
 
+/// Builds the local pub cache directory and returns the path.
+String get _pubCachePath {
+  final pubCachePath = path.join(Directory.current.path, 'local_pub_cache');
```
johnpryan, line 301: Are there any scenarios where this directory would be missing?

johnpryan (approved):

tool/grind.dart
johnpryan: Nevermind, I think `pub ` creates this directory if it doesn't exit...

srawlins: Well, its created on the next line. Pub might create it too though.

PR #790, opened 2021-11-05
Pin firebase packages to current releases

srawlins:


srawlins:
CC @RedBrogdon @puf

We haven't discussed pinning Dart package versions leading up to a period of desired extra stability, but I imagine this is a good thing to do? WDYT?

Currently our scripts will pull whatever latest versions exist, which fit in the version constraints of other dependencies.

tool/grind.dart
```diff
@@ -577,6 +577,20 @@ void updatePubDependencies() async {
       flutterToolPath: sdk.flutterToolPath, channel: _channel);
 }
 
+/// Pinned version numbers for individual packages when extra stability is
+/// desired. Any constraint string is valid (e.g. '^1.2.3' or '>=1.0.0 <3.0.0').
+const Map<String, String> _packagePins = {
```
domesticmouse, line 582: The process for updating these should be documented somewhere

domesticmouse (approved):

domesticmouse:
I'm kinda against this change as it will add ongoing maintenance work. I'd hope latest stable versions work in flutter web, and if they don't it'd be a P0 to fix.

srawlins:
I agree. I'm thinking maybe #791 is a good enough alternative.

My plan with this PR was just that we would pin firebase packages during the 1-2 weeks around an event or announcement or whatever, and then completely unpin them again.

But I definitely take your point; this presumes we know anything about these versions, and changing them within the 1-2 week window may be nontrivial.

domesticmouse:
I'm happy with #791, but it does mean we need to document the pubdeps update step in the DartPad playbook 😎

domesticmouse (changes requested):
Un-approving with reference to my preference to #791 over this change

srawlins:
I'll drop this for now. If we ultimately do want to do this, it'll be easier at this point to create a new PR 😛

PR #792, opened 2021-11-06
Update pub dependencies

domesticmouse:
Update the pub dependencies to bring the firebase deps upto date. This also updates `riverpod`.

/cc @RedBrogdon


srawlins (approved):
❤️ 

Sounds good to me; we can rollback any individual release if we need.

PR #796, opened 2021-11-10
Remove url_launcher

srawlins:
CC @RedBrogdon


domesticmouse (approved):

domesticmouse:
Tests aren't happy for some reason

srawlins:
When I run this locally, I see the following error in Null Safe tests:

```none
/var/folders/gd/9q4y0d095wq9mgmq1wtx8w_w005w15/T/dartpad180si4/lib/bootstrap.dart:1:1: Error: The specified language version is too high. The highest supported language version is 2.14.
            import 'main.dart' as user_code;
            ^
            /var/folders/gd/9q4y0d095wq9mgmq1wtx8w_w005w15/T/dartpad180si4/lib/main.dart:1:1: Error: The specified language version is too high. The highest supported language version is
2.14.
            import 'dart:html';
            ^
```

srawlins:
That may be because I last built the project templates with `FLUTTER_CHANNEL=beta` and then ran the tests without specifying a channel (which defaults to `stable`), and the tests use the local pre-built project templates.

We either need to build the project templates a-fresh for the tests in a sandbox directory (my preference) or short-circuit fail the tests if they detect a mismatch.

srawlins:
I see, removing url_launcher removes our _last_ (only) federated plugin, so now `package:flutter_web_plugins` cannot be imported. Rather than flip-flopping on this support, I think we should just replace the url_launcher package with another one.

domesticmouse (approved):

PR #798, opened 2021-11-11
Make flutter old branch actually n-1.

domesticmouse:


srawlins (approved):
Wow!

domesticmouse:
If only I could get the tests to pass

PR #813, opened 2021-12-30
Rework analysis_server to clear flakiness

srawlins:
The main thrust of this change is to stop using `analysisServer.analysis.onError`, and instead use `analysisServer.analysis.getErrors`.

Events come through `analysisServer.analysis.onError` when adding and removing (I think) overlays, as well as events on `analysisServer.server.onStatus` with `isAnalyzing = false`, so that it is very tricky to handle across asynchronous code which errors and complete analyses are for a given analysis request.

All of the other cleanup (tons of deleted code, simplified warmups, simplified tests) follows from that one change.


domesticmouse (approved):

srawlins:
@scheglov would you mind reviewing this? I think you can help indicate if we are using any specifically outdated APIs here.

test/utils.dart
```diff
@@ -6,7 +6,7 @@ import 'dart:async';
 
 /// Runs [fn] up to three times, catching thrown exceptions, until it returns
 /// normally.
-Future<void> tryWithReruns(FutureOr<void> Function() fn) async {
+Future<void> xtryWithReruns(FutureOr<void> Function() fn) async {
```
scheglov, line 9: Remove it?

scheglov (approved):

test/utils.dart
srawlins: Done

PR #815, opened 2022-01-14
Ensure minimum version numbers for some packages

srawlins:
I've seen non-deterministic behavior from `flutter packages get`, but this minimum version guard seems to fix the issue.


srawlins:
I've run `FLUTTER_CHANNEL=beta grind update-pub-dependencies` a few times with this patch now and it seems to succeed and keep good versions.

domesticmouse (approved):

PR #819, opened 2022-01-29
Dev for development

domesticmouse:
Step 1 of https://github.com/dart-lang/dart-pad/issues/2131


srawlins:
It's confusing to re-use the word, "channel," right? Like this code (and the tests) can be run with FLUTTER_CHANNEL=beta but that still would use `channel: 'stable'` as that channel argument. Hard-coding `channel: 'stable'` in the tests, when they are still run with FLUTTER_CHANNEL=beta and FLUTTER_CHANNEL=old seems confusing. We could call this new parameter, `branch` or something?

WDYT?

domesticmouse:
How about `developmentMode`?

srawlins (approved):
Ah nice, a bool. I love it!

PR #828, opened 2022-02-20
Remove now removed web_sql from allowed imports

parlough:
`dart:web_sql` was removed from the SDK in Dart 2.15.

### DIFF
--- a/lib/src/project.dart
+++ b/lib/src/project.dart
@@ -143,7 +143,6 @@ const Set<String> _allowedDartImports = {
   'dart:svg',
   'dart:web_audio',
   'dart:web_gl',
-  'dart:web_sql',
   'dart:ui',
 };
 


lib/src/project.dart
```diff
@@ -143,7 +143,6 @@ const Set<String> _allowedDartImports = {
   'dart:svg',
   'dart:web_audio',
   'dart:web_gl',
```
domesticmouse, line 145: Do you know the status of `web_gl`? 

Context: https://github.com/dart-lang/dart-pad/issues/2171

domesticmouse (approved):

lib/src/project.dart
parlough: `web_ql` is still supported by browsers and Dart. I commented there on why some web packages may not be showing up.

domesticmouse: Danke

PR #830, opened 2022-02-26
Move google_fonts to only supported by Flutter

parlough:
`package:google_fonts` is only supported by the Flutter SDK according to its [pub.dev entry](https://pub.dev/packages/google_fonts).


parlough:
I'm going to create a new PR which adjusts some other problematic packages.

PR #832, opened 2022-02-26
Remove pedantic from supported packages and imports

parlough:
This would technically be a breaking change(?) as if a snippet imported it previously, they would get an error now.

Closes https://github.com/dart-lang/dart-pad/issues/2180


domesticmouse:
Yup, it'd be breaking. I'll hold off on landing this one until @RedBrogdon and @srawlins have had a chance to comment.

domesticmouse (approved):

srawlins (approved):

PR #833, opened 2022-02-26
Remove now unnecessary pedantic usage

parlough:


domesticmouse:
Broken `old` channel is known. Unsure how I broke it. I suspect it'd be easier if package authors actually specified minimum Flutter versions for their packages.

PR #834, opened 2022-03-03
Use `package_config`

domesticmouse:


RedBrogdon:
Definitely looks easier than parsing everything ourselves. 😄 

@mit-mit is there anything else you'd recommend Brett do here, or are we future-proofed at this point?

mit-mit:
I think this is it, thanks!

PR #846, opened 2022-04-13
Support accepting multiple source files from client to allow for multiple dart files within a Dart-Pad

timmaffett:
This PR allows incoming source file to (optionally) be json object containing a `sources` file set.  
This seemed to be the simplest way to extend the functionality of the dart-services entry points to support having multiple files within a dart-pad project. (see @johnpryan 's https://github.com/dart-lang/dart-pad/issues/2149 and related issues)

No existing use/user should be affected by these changes.

The compiler entry points compile() and compileDDC(), and analysis entry points complete(), getFixes(), getAssists(), dartdoc(), analyze() have been modified to optionally accept a json encoded source filemap in the form of
```json
{
  "files":{
     "filename1" : "sourcecode...",
     ...
     "filenameN": "sourcecode..."
   },
  "active_source_name": "filename1"
}
```
The `files` map also parallels how gists themselves are represented and stored, making preparing this json from the dart-pad side trivial.

Most (all?) of the modified entry points had already created a `sources` map with `{ kDartMain:src }`  to operate on, so this modification simply allows this `sources` map to be passed in the `files` entry of the json.   The existing (forward thinking) internals of operating on the `sources` map allowed the overall changes to be fairly compact and simple. (as well as for example all the output of the analyze based methods already provided source filenames with each fix/info/error/etc).

The `active_source_name` parameter allows the methods which require context (ie. take the editor's `offset` file position within the source) to also have the context of the editor's current file (and the file for which the `offset` refers to).  Any need Location objects are then created with this `active_source_name` instead of being hard coded to `kMainDart`.

I have added a fairly complete testing of the new json parameter.  All existing tests are duplicated using json as the method of passing the source, in addition to testing a small variety of new multiple file json/dart-pad scenarios.  These test the compiler's ability to `import/part/part of`  between the files in the source file set.  I intend on extending these test sets as I complete the dart-pad UI portion of the multiple file support.  The vast majority of the new lines of code are within compiler_test.dart and analysis_server_test.dart.
 
There of course could be lots of other approaches to supporting multiple source file dart-pad's.. I just tried to implement this one in the least impactful way I could see.

Comments, recommendations, criticisms (and outright rejection) of this approach are of course welcome!

other little details:
 - The filenames within the files:{} map are sanitized for safety (no '..', filenames can contain 'package:'/'http:' etc.)
 - If no main.dart file is present (kMainDart within dart-services source) then the file with the main() method is chosen and renamed main.dart.  We would probably already handle ensuring a main.dart on the dart-pad side of things, but just in case..
 - I placed the guts of the json handling is within `common.dart` and the entry point is `getSourcesAndActiveSourceName()` 
 - the formatting entry point was not modified to take a files set, as it did not seem to need to be. (If the ui wants to format all source files it can do them one at a time in turn)
 - All of the files in the source file set must implicitly now be on the 'allowed' list of imports, for this reason I strictly sanitize all the filenames to ensure they do not contain any `'package:'/'dart:'/'http:'/etc.` syntax which could allow an unscrupulous actor to bypass the import restrictions.
  - I originally toyed with the idea of some 'magic' prefix on the json object, but decided to just identify it by looking for a json object that started with the `files` or `active_source_name` entries.  This allows the json string to be directly unpacked and used.  (The limitation of this approach being that the json object must start with the `files` or `active_source_name` entries)


domesticmouse:
Hey @srawlins, if you have time I'd love your eyes on this PR. Thanks!

srawlins:
I can look today or tomorrow.

lib/src/common.dart
```diff
@@ -583,3 +588,149 @@ String stripMatchingQuotes(String str) {
...
+/// Transforms incoming source or json into uniform representation.
+/// [inputSrcOrSources] contains either simple source code
+/// for the only file, or it contains a json object containing
+/// set of source files and the client active source file.
+/// This json object takes form of
+/// `{"files":{"filename1":"sourcecode"...},"active_source_name":"filename from files"}`
+/// returns a SourcesAndActiveSourceName object containing
+/// a uniform representation of the SINGLE or MULTIPLE source
+/// files represented by the original [inputSrcOrSources] string.
+SourcesAndActiveSourceName getSourcesAndActiveSourceName(
```
domesticmouse, line 655: My preference is rather than double encoding information that we add end points to the gRPC bindings that accept the structure we need. We can push the front end and back end with updated gRPC definitions and then the front end can be updated to use the new multi-file capabilities where required.

timmaffett: As I was working on the new gRPC endpoints @domesticmouse requested above I realized that this PR passes the json encoded files list to routines in analysis_servers.dart without decoding.   @srawlins please don't bother looking at this PR until I get the requested changes completed.
   (The test code did not catch this problem in analysis_servers.dat because the code is subsequently checked again for illegal imports at a lower level)

domesticmouse: I've marked this PR as a draft in line with the above comment. Please feel free to mark it ready to review when you are happy to have some eyes on the code. Thanks!

srawlins: 👍 sounds good!

PR #858, opened 2022-04-26
gRPC changes to add alternate API end points which support multiple files - replaces #846

timmaffett:
This PR is intended to replace #846.  This provides back end support for dart-pad having multiple file project/gist capabilities. (see @johnpryan 's https://github.com/dart-lang/dart-pad/issues/2149 and related issues)

This PR adds the following new gRPC end points which support multiple files:
```
/analyzeFiles
/compileFiles
/compileFilesDDC
/completeFiles
/fixesFiles
/assistsFiles
/documentFiles
```

These each accept mutli file versions of the proto3 messages that their single files counter parts accept.
These messages accept a file set map in the form of  { filename1:sourcecode1 .. filenameN:sourcecodeN }
(messages definitions can be found below)

Before this PR some of the existing single file entry points accepted the single file input and wrapped it in a multi file set map like shown above (in this case the map contains only the single `main.dart` entry).   This PR makes it uniform that all single file entry points wrap their single file input into a multi file set map consisting of a `main.dart` entry.
In this way below the entry point both the single and multi api end points share the same code.

  - The other changes included here are the removal of redundant calculation of imports.  This seemed inefficient, and more so with the new capability to compile multiple files.  Imports are now calculated once and passed through to where they are needed.
  - A new, efficient set of `countLines()/countLinesInString()` routines have been added to count the lines of source compiled.  This is done for logging purposes for every compile, previously .split('\n').length() was used, which does needless copying of data.  Again with the capability to compile >1 file it seemed to be the time to do this more efficiently.
  - hashSource() was replaced with _hashSources() - this functions identically for single file compiles and for multiple file set compiles it hashes each file in turn and then hashes the set of hashes.
  - all incoming filenames in the file sets are sanitized against containing up directory navigation (..) and also against containing 'package:', 'dart:', 'http://', 'https://' (the latter could be potentially be used circumvent import restriction checks)
  - If there is no `main.dart` within the fileset then an attempt is made to automatically identify the file containing the main() entry point that the boostrap needs to call.
  - Extensive testing of all of the new entry points has been added.  All test previously executed on the single file entry points are executed against the multiple file entry points, with the addition of a variety of test to exercise the various multiple file specific capabilities. (importing other files and library/part/part of testing, sanitizing, missing main.dart, etc.).  The number of tests have increased from 182 to 306. 
  - Test were added for the new countLines(), countLinesInString() and filename sanitizing routines.
  - Testing of the GET /version end point was added (previously the 'get' test just called the POST /version
 
The net result is that we should now have the added capability of multiple file compiling/analyzing/etc, and in addition having all the existing api end points execute faster and more efficiently as well.

Thank you @domesticmouse and @srawlins for your patience in the long wait for me to complete this.  I lost a couple weeks to the easter break holiday.  I will address any suggestions, issues or corrections that may be needed promptly.
(I also did not know if it was possible to change the branch that an existing PR was based on, and i was lazy and made this new PR instead of updating the existing daft PR #846 - forgive me)

new proto3 messages added for multiple file entry points
```
/// Compile request for a multiple file set
message CompileFilesRequest {
  // The Dart source files set. map of { filename1:sourcecode1 .. filenameN:sourcecodeN }
  map<string, string> files = 1;

  // Return the Dart to JS source map; optional (defaults to false).
  bool returnSourceMap = 2;
}

/// DDC Compile request for a multiple file set
message CompileFilesDDCRequest {
  // The Dart source files set. map of { filename1:sourcecode1 .. filenameN:sourcecodeN }
  map<string, string> files = 1;
}

/// Multiple file set of dart source for analysis, completion, fixes, etc.
message SourceFilesRequest {
  // The Dart source files set. map of { filename1:sourcecode1 .. filenameN:sourcecodeN }
  map<string, string> files = 1;

  // active (within editor) source filename key within files map
  string activeSourceName = 2;

  // The offset within active source file to operate at.
  int32 offset = 3;
}
```


domesticmouse:
@srawlins do you want to also review this PR?

lib/src/analysis_server.dart
```diff
@@ -189,28 +192,35 @@ abstract class AnalysisServerWrapper {
...
-    );
+    return getFixesMulti({kMainDart: src}, Location(kMainDart, offset));
   }
 
   Future<proto.FixesResponse> getFixesMulti(
       Map<String, String> sources, Location location) async {
     final results =
         await _getFixesImpl(sources, location.sourceName, location.offset);
-    final responseFixes = results.fixes.map(_convertAnalysisErrorFix);
+    final responseFixes = results.fixes.map((aef) {
```
domesticmouse, line 202: Nit: please use more expressive parameter names

lib/src/analysis_server.dart
```diff
@@ -313,10 +342,12 @@ abstract class AnalysisServerWrapper {
...
       });
 
+      // get imports if they were not passed in
+      imports ??= getAllImportsForFiles(sources);
+
       // Calculate the imports.
       final packageImports = {
-        for (final source in sources.values)
-          ...getAllImportsFor(source).filterSafePackages(),
+        ...imports!.filterSafePackages(),
```
domesticmouse, line 350: I'm always a tad worried about `!` as it risks bringing the server down. Can we add a `if (imports != null) ...imports.filterSafePackages()` instead?

domesticmouse (commented):
I've started giving this a review, and I'm liking most of what I'm seeing. Just a couple of minor stylistic nits. 

I'll probably be busy until after I/O, but this is looking promising!

srawlins:
Yes I'll try to review today.

lib/src/analysis_server.dart
```diff
@@ -189,28 +192,35 @@ abstract class AnalysisServerWrapper {
...
+  Future<proto.AssistsResponse> getAssistsMulti(
+      Map<String, String> sources, Location location) async {
+    final sourceName = location.sourceName;
+    final results = await _getAssistsImpl(sources, sourceName, location.offset);
+    final fixes =
+        _convertSourceChangesToCandidateFixes(results.assists, sourceName);
     return proto.AssistsResponse()..assists.addAll(fixes);
   }
 
+  /// format the source [src] of the single passed in file.  The [offset] is
```
srawlins, line 221: Nit: Capitalize 'format'.

https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

lib/src/analysis_server.dart
```diff
@@ -267,16 +286,26 @@ abstract class AnalysisServerWrapper {
...
     _logger.fine('analyze: Scheduler queue: ${serverScheduler.queueCount}');
 
     return serverScheduler
         .schedule(ClosureTask<proto.AnalysisResults>(() async {
       sources = _getOverlayMapWithPaths(sources);
       await _loadSources(sources);
-      final errors = (await analysisServer.analysis.getErrors(mainPath)).errors;
+      final List<AnalysisError> errors = [];
+
+      // loop over all files and collect errors -
```
srawlins, line 303: Nit: capitlize 'loop', and finish with a period.

https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

lib/src/analysis_server.dart
```diff
@@ -313,10 +342,12 @@ abstract class AnalysisServerWrapper {
         return a.charStart.compareTo(b.charStart);
       });
 
+      // get imports if they were not passed in
```
srawlins, line 345: Nit: capitalize and period.

https://dart.dev/guides/language/effective-dart/documentation#do-format-comments-like-sentences

lib/src/analysis_server.dart
srawlins: You could also use `...?imports?.filterSafePackages(),`.

srawlins: This is an annoying corner of flow analysis. Normally `imports` would be promoted by the `??= getAllImportsForFiles`, but since the assignment happens in a closure, promotion does not occur. :(

You could also do:

```dart
final packageImports = {
   (imports ?? getAllImportsForFiles(sources)).filterSafePackages(),
}
```

Which doesn't look much better. Maybe extract to a function?

lib/src/common.dart
```diff
@@ -546,6 +564,167 @@ void main(List<String> argv) {
 }
 ''';
 
+/// Code fragments for testing multi file compiling
+///  these are just taken from [sampleCodeFlutterImplicitAnimations] but
```
srawlins, line 568: This line wrapping seems weird. Is this a new sentence here?

lib/src/common.dart
```diff
@@ -546,6 +564,167 @@ void main(List<String> argv) {
...
+              ),
+            ),
+        ],
+      ),
+    );
+  }
+}
+''';
+
+// for multi file 2 part testing using imports
```
srawlins, line 685: Nit: Capitalize and period, here and below.

lib/src/common.dart
```diff
@@ -583,3 +762,135 @@ String stripMatchingQuotes(String str) {
...
+class _SourcesGroupFile {
+  String filename;
+  String content;
+
+  _SourcesGroupFile(this.filename, this.content);
+}
+
+///this RegExp matches a variety of possible `main` function definition formats
+///Like:
+///`Future<void> main(List<String> args) async {`
```
srawlins, line 775: For dartdoc purposes, this should be formatted as a Markdown list, like:

```dart
/// Like:
///
/// * `Future...
/// * `void ...
```

lib/src/common.dart
```diff
@@ -583,3 +762,135 @@ String stripMatchingQuotes(String str) {
...
+    r'''[\s]*(Future)?[\<]?(void)?[\>]?[\s]*main[\s]*\((\s*\w*\s*\<?\s*\w*\s*\>?\s*\w*\s*)?\)\s*(async)?\s*[\{|\=]+''');
+
+/// remove 2 or more '..' and any number of following slashes of / or \ and
+final RegExp sanitizeUpDirectories = RegExp(r'[\.]{2,}[\\\/]*');
+
+/// this will get 'package:', 'dart:', 'http://', 'https://' etc.
+final RegExp sanitizePackageDartHttp = RegExp(r'[\w]*[\:][\/]*');
+
+/// remove anything like package:, dart: http:// from filename
+/// remove runs of more han '.' (no '..' up directories to escape temp)
```
srawlins, line 793: I think there is a typo here. Could you reword this to explain more what this does?

lib/src/common_server_impl.dart
```diff
@@ -154,6 +155,104 @@ class CommonServerImpl {
           .dartdoc(request.source, request.offset, devMode: _sdk.devMode));
   }
 
+  // files map entry points
```
srawlins, line 158: We don't really have a standard for comments like this, grouping a collection of elements. But I do like the practice.
I think if you just separate this from the next line with a blank line, and capitalize and add a period, or a colon, that should look fine.

srawlins (changes requested):
I reviewed a few files quickly, but this will take some time. :) Big change.

lib/src/analysis_server.dart
timmaffett: Again agreed, I was ensuring imports was not null on the line above so didn't check again, but I have added a double check fix which makes me happier.

```dart
  // get imports if they were not passed in
  imports ??= getAllImportsForFiles(sources);

  // Calculate the imports.
  final packageImports = {
    if (imports != null) ...imports!.filterSafePackages(),
  };
```

lib/src/analysis_server.dart
timmaffett: agreed, fixed this to be
```dart
final responseFixes = results.fixes.map((availableAnalysisErrorFixes) {
  return _convertAnalysisErrorFix(availableAnalysisErrorFixes, location.sourceName);
});
```

timmaffett:
@srawlins    I will take a moment and get all of your suggestions addressed as soon as I can today.  I will also comb through my other comments and correct anything I find there as well.

I addressed the null check like this:
```
final packageImports = {
    if (imports != null) ...imports!.filterSafePackages(),
  };
```

but would you prefer the more concise ?

```
final packageImports = {
    ...?imports?.filterSafePackages(),
  };
```

Thank you for taking the time to review the code and offer your suggestions!

domesticmouse:
> but would you prefer the more concise ?
> 
> ```
> final packageImports = {
>     ...?imports?.filterSafePackages(),
>   };
> ```

I have no preference between the two. Given the latter is @srawlins' suggestion, I'd go with that =)

timmaffett:
I have combed through and corrected and clarified all of the comments.  I do apologize for them being in such a state!  

..and I went with @srawlins suggestion:
```
final packageImports = {
    ...?imports?.filterSafePackages(),
  };
```
My clarification question was silly as this is clearly the better choice as it avoids the `!` altogether.

domesticmouse:
Something has gone wrong with the tests by the look of things. We might need to break this PR down into smaller pieces to sneak up on the breakage.

timmaffett:
I saw yesterday that the test said they completed ok but then that test process got to the summary logging and did not exit..
Is that what you are speaking of ? In the log I looked at it had some messages about memory...  They all complete ok/normally in all my local runs. 

I could try commenting out blocks of the test seeing what I can track down - but unfortunately I think you have to trigger all the test running manually for this repo don't you ?

domesticmouse:
I only have to ack tests on your first PR. Once you have a committed PR, the tests just run. That's why I'm asking to break this PR into multiple pieces...

test/compiler_test.dart
```diff
@@ -204,5 +204,231 @@ void main() { print ('foo'); }
...
+      test(
+        'files:{} compileFilesDDC with 3 SANITIZED files using LIBRARY/PART/PART OF',
+        generateCompilerFilesDDCTest({
+          'package:$kMainDart': sampleCode3PartLibraryMain,
+          'dart:discdata.dart': sampleCode3PartDiscDataPartOfTestAnim,
+          'http://various.dart': sampleCode3PartVariousPartOfTestAnim
+        }),
+      );
+
+      // Test renaming the file with the main function ('mymain.dart') to be
```
srawlins, line 325: I don't understand this comment. What "renaming" takes place?

test/compiler_test.dart
```diff
@@ -204,5 +204,231 @@ void main() { print ('foo'); }
...
+          'various.dart': sampleCode3PartVariousPartOfTestAnim,
+          'mymain.dart': sampleCode3PartLibraryMain
+        }),
+      );
+
+      // Two separate files, illegal import in second file, test that
+      // illegal imports within all files are detected.
+      final Map<String, String> filesVar2BadImports = {};
+      const String badImports = '''
+                            import 'package:foo';
```
srawlins, line 340: I think its most idiomatic to indent a multiline string like this at column 0.

test/compiler_test.dart
```diff
@@ -204,5 +204,231 @@ void main() { print ('foo'); }
...
+                            import 'package:bar';
+                            ''';
+      filesVar2BadImports[kMainDart] =
+          sampleCode3PartFlutterImplicitAnimationsImports +
+              "\nimport 'various.dart';\n" +
+              sampleCode3PartFlutterImplicitAnimationsMain;
+      filesVar2BadImports['various.dart'] =
```
srawlins, line 347: I think it would be most idiomatic here to interpolate the strings, in a multiline. Something like

```dart
= '''
$sampleCode3PartFlutterImplicitAnimationsImports
$badImports
...
''';
```

srawlins (dismissed):
I've reviewed some more, but haven't gotten to the tests yet.

lib/src/project.dart
```diff
@@ -179,8 +179,15 @@ String? _packageNameFromPackageUri(String uriString) {
   return uri.pathSegments.first;
 }
 
+/// Goes through imports list and returns list of unsupported imports
+/// Optional [sourcesFileList] contains a list of the source filenames
+/// which are all part of this overal sources file set.
+/// (These are be allowed).  Note that filenames in [sourcesFileList]
+/// were previously sanitized of any 'package:'/etc syntax
```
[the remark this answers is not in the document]
timmaffett: So with the ability to use multiple files we need to import the other files in the files set to be used by other files within the set.  For example:
`   files={'main.dart':'...', 
             'secondfile.dart':'...' };`
within `main.dart` we must do something like `import 'secondfile.dart';`  or  'part 'secondfile.dart';`

The use of `part` is allowed and not checked by the compile process, but the use of `import` places our `secondfile.dart` into the list of imports used by the `main.dart` file.   When we are checking or allowed imports, which primarily is being used to filter out packages which we do not allow/have installed on the server, we now will encounter `secondfile.dart`, so we need this `sourcesFileList` so that the import checking routine can check against that list to know if it is just referencing a file within the source file set.   Without the `sourcesFileList` the file would be rejected for importing a non-allowed import.  Essentially the `sourcesFileList` gets added to our white list of allowed imports.

This is also another reason why I sanitize all of the filenames (keys) of the files={} set map against having any package:/dart:/http: prefixes so that the filenames themselves could not be used to sneak something past the allowed imports check.

Does that make sense ?  I can expand on this in the comment as well if you think that would be better.

test/analysis_server_test.dart
```diff
@@ -247,4 +247,154 @@ void defineTests() {
...
+      await analysisServer.init();
+    });
+
+    tearDown(() => analysisServer.shutdown());
+
+    // Now test multi file 'files:{}' source format.
+    const kMainDart = 'main.dart';
+
+    test('files={} simple_completion', () async {
+      // Just after i.
```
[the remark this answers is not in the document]
timmaffett: This test was copied from the existing single file tests, and for this test the file is:
`
void main() {
  int i = 0;
  i.
}
`

and we are passing in the location after the `.` on the line with `i.` 

I have changed the comment for the single and multi file tests to be
`// Just after `i.` on line 3 of [completionCode]`

test/compiler_test.dart
timmaffett: Our bootstrap.dart file is hard coded to include 'main.dart' (the value const kMainDart within dart-services) and then call main() within this file.  Because of this, I ensure that there is a main.dart within the file set.  If there is no main.dart within the fileset then the file containing the main() entry point must be renamed to `main.dart` .

The alternative would be to modify the bootstrap.dart file - which would be nicer for a few reasons, but requires a few more changes to pass the name of the file set's main() filename along.   The advantages to this would be that the reported errors and such would reference the user's actual filename and not `main.dart	`.  I was imagining handling this on the font end within dart-pad by having dart-pad detect that there was no file called main.dart.  (I think it kind of does this already when loading a gists, it tried for main.dart but then falls back to other possible filenames if no main.dart is found.)
I would prefer to modify bootstrap.dart with the user's filename if you think that is an better solution to renaming... (renaming was just the simplest of possible solutions)

test/compiler_test.dart
timmaffett: That is so much more readable!
I also fixed all of the files I construct in `common.dart` to use interpolation within the multiline strings.  They are much easier to read and evaluate this way.  Somehow it never occurred to me to do that.  Thank you!!

I have implemented all of your valuable suggestions.  Again I sincerely appreciate the time you have taken to review my code.

timmaffett:
ok - I have implemented all of @srawlins suggestions. 

I also ran completed tests suite again locally without issue, but I have made some 'skip:true' modifications to skip some of the multi file test groups in an attempt to see if I can track down what is causing the server test step to not exit after successfully completing test.

lib/src/common.dart
```diff
@@ -556,6 +759,7 @@ class Lines {
     }
   }
 
+  /// Return the 0-based line number.
```
domesticmouse, line 762: Duplicate?

domesticmouse (changes requested):

domesticmouse:
> @domesticmouse - I am still accessing what could be broken out of this PR to make a smaller PR to commit.  In the mean time I will keep checking back to check the test results once you have had a chance to ack the tests for these changes.

For the purposes of being able to debug this PR, adding a space to the README would be sufficient.

test/compiler_test.dart
```diff
@@ -374,7 +374,7 @@ void main() { print ('foo'); }
           expect(result.problems[1].message,
               equals('unsupported import: package:bar'));
         });
-      },skip:true); //TIM TRACKING
+      }, skip: true); //TIM TRACKING
```
domesticmouse, line 377: I'm curious about who TIM is, and why we are tracking him.

timmaffett: :smile:  i have just been putting in that comment so I can easily find all the spots I have changed things to skip tests.

I am getting closer, I have tracked down far enough to where I can get everything the finish cleanly, now I am adding back in tests so I can get to the guilty party.. and then the fun begins - lol.

I will get all the code cleaned up and returned to it's previous state once I have got it. 🤞

timmaffett:
OK @domesticmouse - Got it! 🥳 All test are working and exiting cleanly now. 

There was a mismatched closing of the group() wrapper group test for Null Unsafe/Safe compiler tests.  This resulted in the setUpAll() and tearDownAll() *not* being wrapped around all of the tests.
The compiler variable was declared at the top of the function `late Compiler compiler;` - so when the later tests (which were incorrectly not within the group with the setupAll()/tearDownAll()) executed, somewhere along the async journey the tearDownAll() would finally be called, and from that point on the `compiler` variable was junk - and the outsider test continued using it and :bomb: (the memory somewhere was probably getting hashed.  When executed locally it all probably happened fast enough that it was ok, or the memory didn't get GC'ed, etc.  It's weird the compiler calls still worked and the test passed!?..lol) 

In java I would always have done a `compiler=null;` after the `compiler.dispose();` , just for good measure, to help catch weird things like this.  Here that of course can't be done as is because of null safety - and declaring it `late Compiler? compiler;` seems messy just to catch a random situation like this..
   
```
late Compiler compiler;

...

for (final nullSafety in [false, true]) {
    group('Null ${nullSafety ? 'Safe' : 'Unsafe'} Compiler', () {
      setUpAll(() async {
        final channel =
            Platform.environment['FLUTTER_CHANNEL'] ?? stableChannel;
        compiler = Compiler(Sdk.create(channel));
        await compiler.warmup();
      });

      tearDownAll(() async {
        await compiler.dispose();
      });

  ...many tests...

  }); // group gets closed

 ...more test  HERE, mistakenly not part of group because closed above..  :bomb:
     THESE clobbered memory somewhere after tearDownAll() was called
     at end of group because compiler var got compiler.dispose()'ed and
     it was still being used..

} // end for loop

```

timmaffett:
@domesticmouse / @srawlins  You know i got to thinking that this for loop for nullSafety in `compiler_test.dart` must be a legacy hold out, after checking it's value is never used to configure the compilers (indeed `compile.dart` is now hard wired to always have null-safety on) - so this for loop should really be removed and the group comment changed to 'Null Safe'..
shouldn't it - unless i am missing something... :wink: ?

```
for (final nullSafety in [false, true]) {
    group('Null ${nullSafety ? 'Safe' : 'Unsafe'} Compiler', () {
    ..
    });
}
```

domesticmouse:
Maybe move the compiler declaration to a closer lexical scope and rely on capture in the closures?

timmaffett:
> Maybe move the compiler declaration to a closer lexical scope and rely on capture in the closures?

Done.   That will prevent something similar from happening again.

domesticmouse:
And yes, null safety was a duplication for the old state of having both null safe and null unsafe servers. Feel free to simplify and sanitise as required. 

Might be easier to do in a separate PR 🤓

timmaffett:
@domesticmouse - I made a separate PR for that change as requested.

domesticmouse:
Just to manage expectations, I'm looking to hold dart services where it is until I/O, as we will be part of the show. 

Providing Sam and I give the thumbs up on code reviews, I'd like to land this PR in the week after I/O. 

Thoughts?

timmaffett:
@domesticmouse Thanks for the update, I appreciate it.  That was absolutely my expectations.  I figured you guys would be busy getting ready for I/O.
   My next PR for dart-services should be the Github OAuth end points as a separate shelf handler (I know you mentioned you could take care of it, but I figured you are plenty busy with other things).  That is a fairly small/simple one.

 I was out of the office today and tomorrow, but I am just wrapping up a tangent I went of off on over the weekend (while waiting for the github workflows to hang.. lol) - support for Mermaid diagrams by the markdown package that @srawlins ponders over at the end of https://github.com/dart-lang/markdown/issues/228 .  They are pretty slick. 

  Feel free to point me in the direction of anything that you would like to see done (or even prototyped).

lib/src/project.dart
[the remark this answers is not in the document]
srawlins: I think I understand the set of files. I just don't understand the text about "filenames ... were previously sanitized ..."

Do you mean previous to this PR, or that `sourcesFileList` must be sanitized before being passed to `getUnsupportedImports`?

test/compiler_test.dart
srawlins: Ah interesting. That makes sense.

domesticmouse:
I think this question is for you @RedBrogdon =)

> Feel free to point me in the direction of anything that you would like to see done (or even prototyped).

lib/src/project.dart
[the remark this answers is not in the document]
timmaffett: I mean previous to being passed to getUnsupportedImports().  I have expanded on the comment in the source code to hopefully make this more clear.

lib/src/project.dart
```diff
@@ -178,8 +178,15 @@ String? _packageNameFromPackageUri(String uriString) {
   return uri.pathSegments.first;
 }
 
+/// Goes through imports list and returns list of unsupported imports
+/// Optional [sourcesFileList] contains a list of the source filenames
+/// which are all part of this overall sources file set.
+/// (These are to be allowed).  Note that filenames in [sourcesFileList]
+/// were previously sanitized of any 'package:'/etc syntax
```
srawlins: Cool. This all makes perfect sense and is a very good design.

The one thing I would change is the tone, or the intent of this comment to be aimed at a _user_ of the code. So rather than "were previously sanitized", I think this should be phrased as "must be sanitized".

The [text in Effective Dart](https://dart.dev/guides/language/effective-dart/documentation#do-use--doc-comments-to-document-members-and-types) sort of backs up that doc comments are used to generate documentation, which would be more for the user, and not so much for the maintainer.

timmaffett: I will make that change.  It makes perfect sense, and I was indeed targeting that comment more at the PR evaluator than the end user.
It now reads:
```
/// Goes through imports list and returns list of unsupported imports.
/// Optional [sourcesFileList] contains a list of the source filenames
/// which are all part of this overall sources file set (these are to
/// be allowed).
/// Note: The filenames in [sourcesFileList] were sanitized of any
/// 'package:'/etc syntax as the file set arrives from the endpoint,
/// and before being passed to [getUnsuppotedImports].
/// This is done so the list can't be used to bypass unsupported imports.
/// The function [sanitizeAndCheckFilenames()] was used to sanitize the
/// filenames.
```

flutter-sdk-version.yaml
```diff
@@ -7,14 +7,14 @@
 
 flutter_sdk:
   stable:
-    flutter_version: 2.10.5
```
domesticmouse, line 10: I think this file is out of date, and should melt away once latest is merged / rebased

domesticmouse (approved):
Please update, this basically LGTM once we sort out any version drift inconsistencies

domesticmouse:
Do you have time and a desire to review this @srawlins?

domesticmouse:
Actually the updated linter is [throwing up a hairball](https://github.com/dart-lang/dart-services/runs/6701934751?check_suite_focus=true#step:7:9). Please update

timmaffett:
Ok @domesticmouse those linter issues have been fixed.

PR #861, opened 2022-04-30
add FLUTTER_CHANNEL="stable" prefix to the 'grind test' command 

timmaffett:
add `FLUTTER_CHANNEL="stable"` to prefix `grind test`, as this is `grind test` requires FLUTTER_CHANNEL to be defined also.


timmaffett:
@domesticmouse - I know this is a little more than adding a space.. but when I went to do that I thought this might actually be a useful contribution.

domesticmouse (approved):

domesticmouse:
I'll land once tests turn green

PR #862, opened 2022-05-01
Remove obsolete loop over null safe/unsafe compiler in test and compiler var into group() scope

timmaffett:
This PR removes the obsolete loop over nullSafety true/false - as the variable was never used and the same tests were repeated with no differences.

Changes:
  - The `compiler` variable is moved within group() scope
  - and helper function `generateCompilerDDCTest` is moved within the group() scope (where setupAllI()/tearDownAll() are going to create and dispose() of the `compiler` variable it operares on).
  - the `for (final nullSafety in [false, true]) {` loop is removed
  - the group() name is changed to hard coded '(Always) Null Safe Compiler'

(That it but the diff looks like a lot because it's lost sync and the change in indenting on everything didn't help. )


domesticmouse:
Looks like the branch is out of date, wanna bring it up to spec?

test/compiler_test.dart
```diff
@@ -14,195 +14,190 @@ import 'package:test/test.dart';
...
+      expect(result.problems.length, 1);
+      expect(result.problems[0].toString(), contains('Error: Expected'));
+    });
+
+    test('good import', () async {
+      const code = '''
 import 'dart:html';
 
 void main() {
-  var count = querySelector('#count');
```
domesticmouse, line 149: Revert space deletion please

domesticmouse (approved):
Minor nit, and bring branch up to date

test/compiler_test.dart
timmaffett: sorry I didn't catch that. I un-indented that huge code block and that inadvertently got caught up in it.

Branch up to date and indenting fixed.

domesticmouse (approved):

PR #866, opened 2022-05-10
Adds the two GitHub OAuth endpoints to enable GitHub auth for dart-pad.

timmaffett:
This PR adds the a handler for the 2 GitHub OAuth endpoints used by https://github.com/dart-lang/dart-pad/pull/2227 .
The [dart-pad side ](https://github.com/dart-lang/dart-pad/pull/2227) PR has also been modified to assume these end points exist in the dart-services it is using.

To run and test locally you can pull each corresponding PR (dart-services and dart-pad side),  create your github oauth app id/secret, set the environmental vars for dart-services and grind serve, and you will have github oauth running locally.

All of the configuration parameters for either `server_cloud_run` or `server_dev` are retrieved from environmental variables.
The variable names are (with example for local debugging):
PK_GITHUB_OAUTH_CLIENT_ID=XXXXX
PK_GITHUB_OAUTH_CLIENT_SECRET=XXXXX
K_GITHUB_OAUTH_AUTH_RETURN_URL=http://localhost:8082/github_oauth_authorized
K_GITHUB_OAUTH_RETURN_TO_APP_URL=http://localhost:8000/index.html

The 2 variables that start with PK_xxxxxx do so because all K_xxx environmental variables are dumped to the log and these PK_xxx should be secrets.  The PK_xxx values will be determined when the OAuth accounts are set up on GitHub.  This can be done easily. 
To setup GitHub OAuth App go to https://github.com/settings/developers

For local debugging one would need to create an OAuth App with the values shown in the K_xxx environmental variables above for the corresponding OAuth settings. (walkthrough below)

----------------------
Implementation details

 * <s>Hive is used for tracking the random states of authorizations in progress, this could be done in memory in something as simple Map<String,int> (which is all the hive is) </s>  Redis is now used for all caching (and InMemoryCache() for local debugging)
 * Encryption is used to protect the github authorization token being set back to dart-pad.  This may be overkill and could be something as simple as XOR (I suppose).  I wanted to error on the side of being more secure.
 * The requirement for the encryption package could be removed if the method of encrypting is changed,  the decryption side of things in dart-pad would need to be changed accordingly.
 
-----------------------
(For the actual servers see below)

GitHub OAuth apps will need to be created under 'dart-lang' (or which every account is chosen) for *each* of the dart-services urls where dart-pad might be using.   (It takes only a minute or two)
The `Authorization callback URL`s corresponding to each will be (unless changed in code)
 * https://stable.api.dartpad.dev/github_oauth_authorized
 * https://beta.api.dartpad.dev/github_oauth_authorized
 * https://dev.api.dartpad.dev/github_oauth_authorized
 * https://old.api.dartpad.dev/github_oauth_authorized

The Homepage URL for all of them would be https://dartpad.dev/?  

It would be possible to always use https://stable.api.dartpad.dev (for example) but the code would have to be changed to not use dartServices.rootUrl but instead a hard coded url.

------------------

-To setup GitHub OAuth App go to https://github.com/settings/developers
1) Select 'OAuth Apps' on left side of screen.
2) Select 'New OAuth App' button on upper left
3) `Application Name` => 'Dart-Pad GitHub Gist Auth Endpoint'  (or whatever..)
4) `Homepage URL` - `http://localhost:8000` (or this can be the server/homepage where your dartpad runs)
5) `Authorization callback URL` - This is the IMPORTANT one - this will be
  `http://localhost:8082/github_oauth_authorized` for the OAuth app used for testing locally.  (or if you change the port modify accordingly)
6) Do NOT check 'Enable Device Flow'

You will then need to copy the Client ID provided and choose to generate a client secret and copy that as well.

--------------------


domesticmouse:
> * Hive is used for tracking the random states of authorizations in progress, this could be done in memory in something as simple Map<String,int> (which is all the hive is)

This won't work. We run between 10 and 50 back end servers, depending on load. We use Redis as the in-memory data store that needs to be communal between the running instances. I'd suggest extending the caching infra to handle this requirement and rip Hive out.

timmaffett:
I have switch to using the Redis caching (or InMemoryCache() for services_dev.dart).
Simplified things also because I can use the expiration feature to not have to worry about cleaning abandoned states.

pubspec.yaml
```diff
@@ -20,7 +21,7 @@ dependencies:
   shelf: ^1.0.0
   shelf_router: ^1.0.0
   yaml: ^3.0.0
-
+
```
domesticmouse, line 24: Nit, drop trailing whitespace plz =)

lib/src/github_oauth_handler.dart
```diff
@@ -0,0 +1,404 @@
+// Copyright (c) 2022, the Dart project authors. Please see the AUTHORS file
```
domesticmouse, line 1: Quick question, have you added yourself to the AUTHORS file?

lib/src/github_oauth_handler.dart
```diff
@@ -0,0 +1,404 @@
...
+
+final Logger _logger = Logger('github_oauth_handler');
+
+class GitHubOAuthHandler {
+  /// Entry point called from Dart-Pad to initiate a GitHub OAuth Token request.
+  /// NOTE: any change to [kEntryPointGitHubOAuthInitiate] must also be
+  /// reflected in changing value in dart-pad\lib\github.dart's
+  /// [GitHubUIController.kEntryPointGitHubOAuthInitiate] to
+  /// match.
+  static const kEntryPointGitHubOAuthInitiate = 'github_oauth_initiate';
```
domesticmouse, line 38: Please don't use prefix letters: https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters

lib/src/github_oauth_handler.dart
```diff
@@ -0,0 +1,404 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+///
+/// This presents an API for initiating OAuth requests to github and then
+/// redirecting back to the calling Dart-Pad application.
+///
+/// A OAuth application must be registered with GitHub and if the
+/// initFromEnvironmentalVars() initializer is used then the GitHub client
+/// id and client secret should be stored in the
+/// PK_GITHUB_OAUTH_CLIENT_ID and PK_GITHUB_OAUTH_CLIENT_SECRET
```
domesticmouse, line 12: I've yet to look into it, but I think we will need to integrate this with Cloud Run using https://cloud.google.com/run/docs/configuring/secrets

How does this code cope with running without these environment vars set? I'm happy if it just fails in a polite way as this code will land and be in prod before we sort out all the keys.

domesticmouse (changes requested):
This basically LGTM, but with a few minor changes required. This will also need to wait for a security review et al.

lib/src/github_oauth_handler.dart
timmaffett: There are 2 environmental variabkes that I expect to come from cloud run secrets - 
    PK_GITHUB_OAUTH_CLIENT_ID
and 
    PK_GITHUB_OAUTH_CLIENT_SECRET

I prefix these with 'PK_' instead of 'K_' to denote that they are private constants that need to be stored as secrets.  
   If these variables are missing then it is noted in the log and the OAuth system does no further initialization,  deactivating itself.  (The 2 OAuth routes are NOT added to the router)

Additionally I will add that the PK_xxx environmental variables, when present, are masked when logged and only show the last 4 characters to aid in verifying that the correct values are being used.

lib/src/github_oauth_handler.dart
timmaffett: I have removed the 'k' prefix - I typically adhere to the not using prefix letters, but these were only prefixed to indicate that they get set from the corresponding environmental variables:
  K_GITHUB_OAUTH_AUTH_RETURN_URL
and
  K_GITHUB_OAUTH_RETURN_TO_APP_URL

(and the usage of 'K_' prefix on the environmental variables I just copied from other environmental variables that dart services uses)

lib/src/github_oauth_handler.dart
timmaffett: done.

lib/src/github_oauth_handler.dart
domesticmouse: Ahh, that makes sense.

domesticmouse:
Yup, the cache is shared between all four flavours of compilation backend.

AUTHORS
```diff
@@ -5,3 +5,4 @@
 
 Google Inc.
 George Moschovitis <george.moschovitis@gmail.com>
+Tim Maffett <timmaffett@gmail.com>
```
kevmoo, line 8:
```suggestion
Tim Maffett <timmaffett@gmail.com>

```

kevmoo: New lines at the end of file is great!

domesticmouse (approved):

domesticmouse:
This PR is now serving on all four endpoints, I believe configured with appropriate Github OAuth client and secrets.

PR #867, opened 2022-05-11
Make CommonServerApi.router getter behave nicely and not return a new router each time it is accessed.

timmaffett:
The existing implementation of the `CommonServerApi.router` getter creates a new router object *each time* it is accessed.  This seems contrary to how a getter should behave and is counter intuitive to the behavior of a getter.  

I ran into this problem in problem in my PR #866 when attempting to add routes to the router.  Initially it seemed to not work and it was quickly apparent what the problem was, I was adding routes to a different router then the one subsequently being created when the getter was accessed later (when it was used in the `addHandler()` call).   
For that PR solved the problem by accessing the `router` getter only once but it nagged at me so I created this simple pr to fix it's behavior.


lib/src/common_server_api.dart
```diff
@@ -107,7 +107,10 @@ class CommonServerApi {
           decodeFromProto: (bytes) => proto.VersionRequest.fromBuffer(bytes),
           transform: _impl.version);
 
-  Router get router => _$CommonServerApiRouter(this);
+  Router? _router;
+
+  /// Return the router, creating it if we have not done so already.
+  Router get router => (_router ??= _$CommonServerApiRouter(this));
```
parlough, line 113: I'm not sure if there was a reason it was constructed each time before, but you can use `late final` to make this type of  structure a bit easier to read:

```suggestion
  /// The (lazily-constructed) router.
  late final Router router = _$CommonServerApiRouter(this);
```

timmaffett: Yes - that is much better.  I have corrected that.

The existing code accesses the getter only once, when the server is being created in `services_cloud_run.dart`  or `services_dev.dart`.

Previously there was no occasion to access it again.  It seems to me that if the previous behavior was desired then this should be a method call with an appropriate name and not a getter.

domesticmouse (approved):

domesticmouse:
Danke!

PR #869, opened 2022-05-12
Dart 2.17.0

domesticmouse:


example/pubspec.lock
```diff
@@ -429,4 +429,4 @@ packages:
     source: hosted
     version: "3.1.0"
 sdks:
-  dart: ">=2.16.0-100.0.dev <3.0.0"
+  dart: ">=2.17.0-206.0.dev <3.0.0"
```
johnpryan, line 432: 2.17.0?

domesticmouse: That's the `pubspec.lock` file. It's whatever `dart pub` chooses to make it.

johnpryan (approved):

PR #874, opened 2022-06-16
Flutter 3.1.0-9.0.pre

domesticmouse:


domesticmouse:
Hey @srawlins, can you give guidance on what the Web bootstrap should now be? Flutter 3.1.0-9.0.pre has broken the old bootstrap file that expected `generated_plugin_registrant.dart` and I'm unsure what the bootstrap should now look like...

srawlins:
Do you have CLs where they've been changing what the bootstrap file is going to be generated as? The CI failures don't look super helpful, but can we get the server logs for those requests?

domesticmouse:
There are no server logs as this hasn't been landed yet. Once this PR lands, the beta server will automatically adopt it, and start failing requests that rely on plugins. I'm guessing that means most of the Firebase packages, for starters.

The before and after on the build method looks like this:

  - https://github.com/flutter/flutter/blob/169020719bc5882e746b836629721644633b6c8a/packages/flutter_tools/lib/src/build_system/targets/web.dart#L137
  - https://github.com/flutter/flutter/blob/main/packages/flutter_tools/lib/src/build_system/targets/web.dart#L176

The delta between the two is too large to get GitHub to generate a visual diff.

srawlins:
For the server logs, I just mean the server during the test. "Expected 200, Actual 400" isn't super helpful. So we'd have to add debugging or something to see why it's so angry.

domesticmouse:
With just the version of the Dart SDK bumped to `3.1.0-9.0.pre` and feeding the DDC compiler the following:

```dart
import 'package:flutter/material.dart';

void main() {
  for (int i = 0; i < 4; i++) {
    print('hello ${i}');
  }
}
```

We see the following error:

```console
Error: Error when reading '/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadU0yzQJ/.dart_tool/package_config.json': No such file or directory
Error: Couldn't resolve the package 'flutter_web_plugins' in 'package:flutter_web_plugins/flutter_web_plugins.dart'.
/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadU0yzQJ/lib/bootstrap.dart:2:8: Error: Not found: 'package:flutter_web_plugins/flutter_web_plugins.dart'
import 'package:flutter_web_plugins/flutter_web_plugins.dart';
       ^
/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadU0yzQJ/lib/bootstrap.dart:3:8: Error: Error when reading '/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadU0yzQJ/lib/generated_plugin_registrant.dart': No such file or directory
import 'generated_plugin_registrant.dart';
       ^
/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadU0yzQJ/lib/bootstrap.dart:7:19: Error: Undefined name 'webPluginRegistrar'.
  registerPlugins(webPluginRegistrar);
                  ^^^^^^^^^^^^^^^^^^
/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadU0yzQJ/lib/bootstrap.dart:7:3: Error: Method not found: 'registerPlugins'.
  registerPlugins(webPluginRegistrar);
  ^^^^^^^^^^^^^^^
```

With the contents of this PR, the returned error is now:

```console
/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadHQdH4t/lib/bootstrap.dart:5:8: Error: Error when reading '/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadHQdH4t/lib/web_plugin_registrant.dart': No such file or directory
import 'web_plugin_registrant.dart' as pluginRegistrant;
       ^
/var/folders/5y/4dz9m6091ms_72g3zk9pvjn4003nrc/T/dartpadHQdH4t/lib/bootstrap.dart:19:24: Error: Method not found: 'registerPlugins'.
      pluginRegistrant.registerPlugins();
                       ^^^^^^^^^^^^^^^
```

srawlins:
Maybe @ditman can help out here. I'm not sure exactly what changed with https://github.com/flutter/flutter/pull/102185, but our current process for running a Flutter Web app is:

1. User clicks 'run' and the browser sends the user script to the server.
2. We create a little temp project directory with the user script as `lib/main.dart`, I think a `web/index.html` maybe, and a pubspec with entries that depend on what the user script contained ("Oh it looks like they need firebase, I'll add that").
3. We run `flutter packages get` in that dir, [here](https://github.com/dart-lang/dart-services/blob/master/lib/src/project_creator.dart#L100).
4. We expect `lib/generated_plugin_registrant.dart` has been created.
5. We create our own `lib/bootstrap.dart` file which refers to that `lib/generated_plugin_registrant.dart` file [here](https://github.com/dart-lang/dart-services/blob/master/lib/src/compiler.dart#L175).
6. We compile the user's app by directly calling `dartdevc` (DDC), [here](https://github.com/dart-lang/dart-services/blob/master/lib/src/compiler.dart#L185..L206).

I suspect that we create our own `bootstrap.dart` file because we don't run `flutter build`. I don't think I know why we don't run `flutter build`. Something about DDC? require.js? But there are ways to run `flutter build` which use DDC.

So I imagine `flutter build` is creating one or more files that we would now need to generate ourselves.

ditman:
> 4. We expect `lib/generated_plugin_registrant.dart` has been created.

Ahhh, that is no longer true, the file gets created by `flutter build` / `flutter run` at the moment it's needed to compile the app!

---

There was also a small change in the contents of the plugin registrant and how it's generated that might explain the errors that @domesticmouse is seeing:

So when there are no plugins we generate a file that has just a noop `registerPlugins` method:

* https://github.com/flutter/flutter/blob/1af8cc118390f3b0e53164b150ccc69339e9329c/packages/flutter_tools/lib/src/flutter_plugins.dart#L553-L562

And when there ARE plugins, we assume we can fallback to the webPluginRegistrar, but the default entrypoint always calls this as `registerPlugins()` now (so we save on a dependency on `flutter_web_plugins` on the bootstrap code itself):

* https://github.com/flutter/flutter/blob/1af8cc118390f3b0e53164b150ccc69339e9329c/packages/flutter_tools/lib/src/flutter_plugins.dart#L577-L578

This is the new bootstrap code:

* https://github.com/flutter/flutter/blob/1af8cc118390f3b0e53164b150ccc69339e9329c/packages/flutter_tools/lib/src/web/file_generators/main_dart.dart#L31-L43

Note that the bootstrap just defines the `runApp` and `registerPlugins` methods, which then get called by the programmer (if they choose to use the new initialization, from their JavaScript, [example](https://github.com/flutter/gallery/blob/main/web/index.html#L85-L101)) or by the engine "automatically", as it did before this change (so you wouldn't need any changes to your old index.html if it's being generated somewhere). Here's the [source for `webOnlyWarmupEngine`](https://github.com/flutter/engine/blob/cfbfb0c1776c91a5442b2c7591bf383ec9e0f56f/lib/web_ui/lib/initialization.dart#L35-L63), for reference.

---

So I guess, your flutter "bootstrap.dart" [here](https://github.com/dart-lang/dart-services/blob/8c77a6e208ab918b3fc89959e77867bfd1829400/lib/src/common.dart#L21-L23) needs to be rewritten around `ui.webOnlyWarmupEngine`.

The plugin registrant seems harder, I guess you guys don't want to `flutter build` because it slows down execution compared to DDC?

domesticmouse:
> The plugin registrant seems harder, I guess you guys don't want to `flutter build` because it slows down execution compared to DDC?

My understanding is that with DDC we can do a chunk of work upfront (e.g. compiling Flutter and Dart SDK to JS) and then have a quick per-request compilation. We want to be able to compile flutter code in sub second timeframes to prevent blowing out the number of serving machines we need. It's also nice for users to have as quick as possible response to the compile-and-run button.

Moving from DDC to full compilation would be a major undertaking, invalidating several pieces of our infrastructure, and unlikely to be doable before the next stable release.

ditman:
So, we should restore the `generated_plugin_registrant.dart` on `flutter pub get`?

We could put it somewhere in .dart_tool for you to move around as needed? The goal of the change above was to *not* create it the users' working copy (inside `lib`)

ditman:
Created this issue: https://github.com/flutter/flutter/issues/106765 /cc @yjbanov

domesticmouse:
> So, we should restore the `generated_plugin_registrant.dart` on `flutter pub get`?
> 
> 
> 
> We could put it somewhere in .dart_tool for you to move around as needed? The goal of the change above was to *not* create it the users' working copy (inside `lib`)

This works for me!

ditman:
A fix for this has landed in `master`, and a cherry-pick for it has been requested for `beta`.

ditman:
@domesticmouse the generation of the flutter plugin registrant on `flutter pub get` has landed in `flutter beta`: `3.3.0-0.0.pre`. 

Apologies for the delay, my CP request was skipped in favor of the normal beta flow.

PR #916, opened 2022-08-31
Create `master`-ish channel

domesticmouse:
The intent with this PR is to create a new hidden channel for DartPad that contains a weekly updated copy of `master` to better support @gspencergoog's master API docs.

Note, I had to de-optimise the Flutter checkout, as the previous approach left a broken Flutter checkout when using it on channel `master`, for some unknown reason.


srawlins (approved):
Cool!

PR #935, opened 2022-10-04
Drop AppEngine text from README

domesticmouse:
`dart-services` no longer supports AppEngine as a deployment target


domesticmouse:
PTAL @RedBrogdon

RedBrogdon (approved):

PR #939, opened 2022-10-08
Allow current library (empty) import

parlough:
Fixes https://github.com/dart-lang/dart-pad/issues/2384

### DIFF
--- a/test/flutter_web_test.dart
+++ b/test/flutter_web_test.dart
@@ -35,6 +35,11 @@ void defineTests() {
       expect(usesFlutterWeb({packageFlutterImport}, devMode: false), isTrue);
     });
 
+    test('getUnsupportedImport allows current library import', () {
+      expect(getUnsupportedImports([_FakeImportDirective('')], devMode: false),
+          isEmpty);
+    });
+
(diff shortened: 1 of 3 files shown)


test/flutter_web_test.dart
```diff
@@ -35,6 +35,11 @@ void defineTests() {
       expect(usesFlutterWeb({packageFlutterImport}, devMode: false), isTrue);
     });
 
+    test('getUnsupportedImport allows current library import', () {
+      expect(getUnsupportedImports([_FakeImportDirective('')], devMode: false),
+          isEmpty);
+    });
+
     test('getUnsupportedImport allows dart:html', () {
```
domesticmouse, line 43: Can we add a test for compiling the source? I'm not sure the above test is sufficient to explain the intent of the change.

```dart
import "" as foo;

int bar = 2;

void main() {
  print(foo.bar);
}
```

parlough: Good idea, thanks.

domesticmouse (approved):

jonahwilliams:
Awesome, thank you @parlough !

PR #955, opened 2022-10-28
Bump Firebase versions by removing version pinning

domesticmouse:
This is a break fix for `firebase_core` 1.24.0.


lib/src/project_creator.dart
```diff
@@ -166,9 +166,7 @@ linter:
 Map<String, String> packageVersionConstraints({required bool oldChannel}) {
```
RedBrogdon, line 166: Is this method still needed?

RedBrogdon (approved):

lib/src/project_creator.dart
domesticmouse: Currently no. But I'm leaving it in place for the next time Firebase breaks the old channel.

PR #963, opened 2022-11-09
Migrate logging to `gcp`

domesticmouse:


kevmoo (commented):
Sooo...the logging logic only does "magic" when used WITHIN a request that has been wrapped with the "right logic".

I think you should hold off here until I figure out a plan w/ that.

domesticmouse:
> I think you should hold off here until I figure out a plan w/ that.

Any progress @kevmoo?

kevmoo:
Baby steps, but I'll likely have to wait until after January events.

I'd mark this as a draft and leave it for a bit...

PR #964, opened 2022-11-09
blast_repo fixes

kevmoo:
Dependabot
GitHub Action


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# Dependabot configuration file.
```
devoncarew, line 1: Note that there's an existing https://github.com/dart-lang/dart-services/blob/master/.github/dependabot.yaml file in this repo; we should update the tool to handle this.

kevmoo: Yup! Well close out this PR and re-evaluate!

PR #972, opened 2022-12-02
Add Patterns and Records experiments to master channel

domesticmouse:
This adds the patterns and records experiment flag to only the DDC compiler on the master channel instance.

Do we also need to add flags to the Analyzer instance?

### DIFF
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -194,6 +194,10 @@ class Compiler {
         ...['--module-name', 'dartpad_main'],
         '--enable-asserts',
         '--sound-null-safety',
+        if (_sdk.masterChannel) ...[
+          '--enable-experiment=patterns',
+          '--enable-experiment=records',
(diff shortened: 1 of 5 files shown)


srawlins:
> Do we also need to add flags to the Analyzer instance?

Yeah. They can be specified in the analysis_options.yaml file:

```yaml
analyzer:
  enable-experiment:
    - records
    - patterns
```

domesticmouse:
> > Do we also need to add flags to the Analyzer instance?
> 
> Yeah. They can be specified in the analysis_options.yaml file:
> 
> ```yaml
> analyzer:
>   enable-experiment:
>     - records
>     - patterns
> ```

Added, for both Dart and Flutter. Note, the Dart2js compiler is explicitly not enabled, so this might be a bit misleading.

srawlins (approved):
Sweet, this looks perfect!

PR #979, opened 2022-12-14
Add experiments to `VersionResponse`

domesticmouse:
fixes: https://github.com/dart-lang/dart-pad/issues/2429

### DIFF
--- a/lib/src/common_server_impl.dart
+++ b/lib/src/common_server_impl.dart
@@ -274,7 +274,14 @@ class CommonServerImpl {
         ..flutterDartVersionFull = _sdk.versionFull
         ..flutterVersion = _sdk.flutterVersion
         ..packageVersions.addAll(packageVersions)
-        ..packageInfo.addAll(packageInfos),
+        ..packageInfo.addAll(packageInfos)
+        ..experiment.addAll(
+          [
+            // TODO(DomesticMouse): update when records and patterns experiments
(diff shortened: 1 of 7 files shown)


lib/src/common_server_impl.dart
```diff
@@ -274,7 +274,14 @@ class CommonServerImpl {
         ..flutterDartVersionFull = _sdk.versionFull
         ..flutterVersion = _sdk.flutterVersion
         ..packageVersions.addAll(packageVersions)
-        ..packageInfo.addAll(packageInfos),
+        ..packageInfo.addAll(packageInfos)
+        ..experiment.addAll(
+          [
+            // TODO(DomesticMouse): update when records and patterns experiments
+            // are modified or removed.
+            if (_sdk.masterChannel) ...['records', 'patterns'],
```
johnpryan, line 282: Do we set these manually when we deploy? Is there a single source of truth somewhere?

domesticmouse: https://github.com/dart-lang/dart-services/pull/972/files

johnpryan: One nitpick would be to keep a top-level list of experiments to avoid repeating ourselves, otherwise LGTM

johnpryan (approved):
LGTM - thanks for the quick turnaround. Feel free to land and I'll update the frontend

PR #995, opened 2023-01-11
Add new dependencies to `dev`

domesticmouse:


lib/src/project.dart
```diff
@@ -77,13 +77,9 @@ const Set<String> firebasePackages = {
 
 /// The set of supported Flutter-oriented packages.
 Set<String> supportedFlutterPackages({required bool devMode}) => {
-      'flutter_bloc',
```
domesticmouse, line 80: This is just sorting the packages, feel free to ignore.

lib/src/project.dart
```diff
@@ -93,19 +89,34 @@ Set<String> supportedFlutterPackages({required bool devMode}) => {
...
       'yaml_edit',
-      'tuple',
-      'animations',
-      'equatable',
-      'matcher',
-      'flutter_map',
-      'petitparser',
-      if (devMode) ...[],
+      if (devMode) ...[
+        'flutter_adaptive_scaffold',
```
domesticmouse, line 109: @johnpryan please notate any packages you think I should remove

tool/pub_dependencies_dev.json
```diff
@@ -44,20 +54,27 @@
...
   "flutter_processing": "0.2.0",
   "flutter_riverpod": "2.1.3",
+  "flutter_spinkit": "5.1.0",
   "forge2d": "0.11.0",
+  "frontend_server_client": "2.1.3",
+  "glob": "2.1.1",
   "go_router": "6.0.1",
   "google_fonts": "3.0.1",
   "hooks_riverpod": "2.1.3",
+  "html": "0.15.1",
```
domesticmouse, line 71: @srawlins looks like `html` is a dependency of something in this batch. Should we open up `html` for anyone to use on DartPad?

srawlins (approved):

lib/src/project.dart
```diff
@@ -93,19 +89,34 @@ Set<String> supportedFlutterPackages({required bool devMode}) => {
...
+        'dartz',
+        'fast_immutable_collections',
+        'timezone',
+        'shared_preferences',
```
johnpryan, line 113: Do we need to verify this is working?

johnpryan (approved):

lib/src/project.dart
domesticmouse: The easiest way to confirm this is to merge this PR and then go back to the original issues and ask each submitter to verify their package is working as intended

PR #1003, opened 2023-01-22
Revert `dev` channel back to Flutter stable

domesticmouse:
I goofed in https://github.com/dart-lang/dart-services/pull/1002 and broke the build for the DartPad `dev` channel. Reverting just that part of the previous change.


mit-mit (approved):

mit-mit:
I'm curious, what is `dev` in the DartPad UI (which has Stable, Beta, Old, and Master)?

domesticmouse:
> I'm curious, what is `dev` in the DartPad UI (which has Stable, Beta, Old, and Master)?

Dev was the old flutter dev channel. When that channel was retired we repurposed it for testing things. So these days the dev channel shadows the stable channel, but it has additional packages. It's also hidden by default in the front end UI.

So these days stable and beta are the matching flutter channels. Old is the previous stable, and master is my best effort at keeping up to date with google3 rolls. Usually I sim for weekly.

PR #1005, opened 2023-01-23
Disable experiments until Dart2JS and DDC support the same flags.

domesticmouse:
The difference in experiment support between Dart2JS (straight Dart dartpads) and DartDevCompiler (Flutter dartpads) is causing too much confusion.

Re-enable experiments once Dart2JS and DartDevCompiler support the same experiment flags.

### DIFF
--- a/lib/src/compiler.dart
+++ b/lib/src/compiler.dart
@@ -194,9 +194,8 @@ class Compiler {
         ...['--module-name', 'dartpad_main'],
         '--enable-asserts',
         '--sound-null-safety',
-        if (_sdk.masterChannel)
-          ..._sdk.experiments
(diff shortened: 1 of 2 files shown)


lib/src/compiler.dart
```diff
@@ -194,9 +194,8 @@ class Compiler {
         ...['--module-name', 'dartpad_main'],
         '--enable-asserts',
         '--sound-null-safety',
-        if (_sdk.masterChannel)
```
domesticmouse, line 197: This guard isn't required, because the experiments are gated in the `Sdk` class.

srawlins (approved):
SG

PR #1006, opened 2023-01-24
Re-enable experiment flags with Dart2JS included

domesticmouse:
Hey Sam, can you confirm that I am adding the experiment flags for Dart2JS in the right spot?


srawlins:
I'm really not sure, @sigmundch could you help us out?

lib/src/compiler.dart
```diff
@@ -88,6 +88,8 @@ class Compiler {
         '--packages=${path.join('.dart_tool', 'package_config.json')}',
         '--sound-null-safety',
         '--enable-asserts',
+        ..._sdk.experiments
+            .map((experiment) => '--enable-experiment=$experiment'),
```
sigmundch, line 92: it appears that dart2js only supports enabling multiple experiments by passing them all to a single flag, instead of passing multiple `--enable-experiment` flags. 

I believe you need to change this to:
```dart
...
'--enable-asserts',
if (_sdk.experiments.isNotEmpty)
  '--enable-experiment=${_sdk.experiments.join(",")}',
'-o',
...
```

domesticmouse: I've updated the code to use a single `--enable-experiment` flag for both Dart2JS and DDC

domesticmouse:
PTAL @johnpryan

johnpryan (approved):

PR #1019, opened 2023-02-09
perf: don't summarize when compiling with DDC

Salakar:
DDC generates a .dill summary file by default which is resource intensive and unnecessary since the file is not even used afterward (in the context of DartPad).

I don't have time to benchmark the before & after results of this change for DartPad sorry, but it should be a significant improvement for the compilation of Flutter apps; turning this off hugely improved compilation times on Zapp (2-3x reduction in compilation time, sometimes even much higher).


domesticmouse:
Thank you @Salakar!

PR #1028, opened 2023-02-22
Move to pkg:dart_flutter_team_lints

kevmoo:
Some non-trivial, but nice cleanup in lib/src/sdk.dart

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,4 +1,4 @@
-include: package:lints/recommended.yaml
+include: package:dart_flutter_team_lints/analysis_options.yaml
 
 analyzer:
   language:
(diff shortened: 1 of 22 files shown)


lib/src/sdk.dart
```diff
@@ -282,6 +273,14 @@ class _DownloadedFlutterSdk {
         .transform<String>(utf8.decoder)
         .listen((string) => stderr.write(string));
 
-    return await process.exitCode;
+    final code = await process.exitCode;
+    if (throwOnError && code != 0) {
```
kevmoo, line 277: this just seems better all around

analysis_options.yaml
```diff
@@ -13,18 +13,11 @@ analyzer:
     - lib/src/protos/**
     - local_pub_cache/**
     - project_templates/**
+  errors:
+    lines_longer_than_80_chars: ignore
 
 linter:
   rules:
-    - always_declare_return_types
```
kevmoo, line 19: Removed duplicative lints

kevmoo:
CC @parlough

tool/fuzz_driver.dart
```diff
@@ -32,9 +32,9 @@ analysis_server.AnalysisServerWrapper? analysisServer;
 
 late comp.Compiler compiler;
 
-var random = Random(0);
-var maxMutations = 2;
-var iterations = 5;
+Random random = Random(0);
```
domesticmouse, line 35: I'm curious why you are adding types here, but losing them elsewhere

domesticmouse (approved):

tool/fuzz_driver.dart
kevmoo: that IS weird. It's just the tooling. I'll look quick.

parlough: I believe this comes from the [`type_annotate_public_apis`](https://dart.dev/tools/linter-rules#type_annotate_public_apis) lint which prefers explicit type annotations on top-level variables.

PR #1035, opened 2023-03-15
Flutter 3.9.0-0.1.pre

domesticmouse:
@RedBrogdon I heard you liked PR review requests =)


lib/src/sdk.dart
```diff
@@ -88,7 +88,7 @@ class Sdk {
     // The Flutter web SDK path changed. The SDK version test here will go
     // through the master, beta, stable, then old waterfall. Well, the last step
     // is the removal of this test, but you get the idea.
-    if (masterChannel) {
+    if (masterChannel || betaChannel) {
```
johnpryan, line 91: I'm guessing we'll delete this once the stable release lands?

johnpryan (approved):

lib/src/sdk.dart
parlough: Or at the latest, when the `old` channel uses the new path.

Tracking in https://github.com/dart-lang/dart-pad/issues/2478

domesticmouse: It'll be deleted in Q3 when `old` gets the updated path.

tool/pub_dependencies_beta.json
```diff
@@ -143,7 +143,7 @@
   "vector_math": "2.1.4",
   "video_player": "2.6.0",
   "video_player_android": "2.4.2",
-  "video_player_avfoundation": "2.4.1",
```
karbonnanroid, line 146: r

domesticmouse: These are not maintained manually, but generated by running pub on a set of packages. If this particular package now needs to be explicitly relied on, I'm curious as to why.

PR #1048, opened 2023-04-05
Flutter 3.7.10

domesticmouse:


srawlins (approved):

parlough (approved):
Looks good to me. Only was concerned about the flame changes breaking PadRacing, but it seems the breaking changes in 1.7.0 shouldn't affect it.

parlough:
Make sense!

PR #1059, opened 2023-04-20
Drop `example`

domesticmouse:
It's basically unused, and it is causing issues on Cloud Build.


johnpryan (approved):
LGTM, but it would be nice to keep the example around

johnpryan:
What's the error coming from Cloud Build look like?

PR #1076, opened 2023-05-10
Flutter-3.10

domesticmouse:


lib/src/sdk.dart
```diff
@@ -95,7 +95,7 @@ class Sdk {
     // The Flutter web SDK path changed. The SDK version test here will go
     // through the master, beta, stable, then old waterfall. Well, the last step
     // is the removal of this test, but you get the idea.
-    if (masterChannel || betaChannel) {
+    if (masterChannel || betaChannel || stableChannel) {
```
johnpryan, line 98: probably dev channel too? Maybe we can take out this if statement now...

johnpryan (approved):

lib/src/sdk.dart
parlough: I believe `master` is `dev`, and I guess `old` still needs the old behavior?

domesticmouse: We can take it out of `old` ... in three months.

parlough:
Actions has really been a pain today :/

domesticmouse:
> Actions have really been a pain today :/

I'm guessing everything broke all at once...

PR #1077, opened 2023-05-10
Fix up `beta` and `old` channel tests

domesticmouse:
Hey @johnpryan, here's the promised `beta` channel test fix up PR.


parlough (approved):
🎸

PR #1099, opened 2023-06-26
Fixes for expansion of `prefer_final_in_for_each`

parlough:
Caught due to recent expansion of the lint rule in https://github.com/dart-lang/linter/commit/f3991273ea6811975b9377ee50430ca8391ab392.


domesticmouse (approved):
I'm all for this change, but I feel it would be better if `prefer_final_in_for_each` was added to the applied lints to make sure it sticks.

parlough:
I believe it already: https://github.com/dart-lang/dart-services/blob/master/analysis_options.yaml#L18

I believe these aren't caught by CI yet since the fix catching these cases landed recently: https://github.com/dart-lang/linter/commit/f3991273ea6811975b9377ee50430ca8391ab392

PR #1103, opened 2023-07-13
Flutter 3.10.6

domesticmouse:


domesticmouse:
> Looks good to me. 
> 
> However, is the plan to switch deployment to the dart-pad repository soon since dart-services also lives there now too? Just want to clarify for myself before they diverge too much. 

I'm keeping this repo up to date with dart and flutter versions while @johnpryan investigates the potential to automate version changes in the dart-pad version. That is my understanding, but John can correct me here if required.

PR #1104, opened 2023-07-14
Drop deprecated fields

domesticmouse:


devoncarew (approved):
Thanks for mirroring my PR here! I'll look to update both repos for future changes while this is still the system-of-record.

